PR #3, opened 2022-07-17
Add pyright as well as tests and infrastructure inspired by bugbear

jakkdl:
I didn't find a way to configure pyright in `setup.py`, opted to use a `pyproject.toml` instead of a `pyrightconfig.json`.

With `pytest-xdist` commented out passing `-n auto` errors out. Commented out the parameter, but probs cleaner to uncomment both lines or remove both.

Could opt to explicitly list files to avoid checking `setup.py` instead of adding the check for `__version__` in it.


jakkdl:
All checks are for stuff I didn't touch in this pull request, opening another one that works on `flake8_trio.py` and adds basic tests.

Zac-HD:
- pyright config lgtm, discovery is less error-prone than explicit lists
- let's leave xdist commented out for now; we'll uncomment once we have ~10 tests and concurrency is worth the overhead
- obvious test failures are obvious, feel free to fix them here or in another PR as you prefer.

jakkdl:
kk, let's just do that here then.
I haven't delved into understanding the testing setup that bugbear uses, so just doing basic tests for now.

flake8_trio.py
```diff
@@ -34,15 +33,19 @@ def is_trio_call(node: ast.AST, *names: str) -> Optional[str]:
...
+
+        # Context manager with no `await` call within
+        calls = [
+            is_trio_call(item.context_expr, "fail_after", "move_on_after")
+            for item in node.items
+        ]
+        if any(calls) and not any(isinstance(x, ast.Await) for x in ast.walk(node)):
+            self.problems.append(
+                (node.lineno, node.col_offset, TRIO100.format(calls.pop()))
+            )
```
Zac-HD, line 48: Actually important to use the `item` attributes as in my previous loop, so that we point at the call to `trio.xxx_after()` if there are multiple clauses e.g.
```python
with (
    trio.move_on_after(10),
    httpx.AsyncContext() as client,
): ...
```

Zac-HD (commented):
LGTM with a comment below - I would recommend flake8-bugbear testing, it amounts to "write a Python file, then get results for that file, and assert that they're the expected results".  Just a bit easier to read than embedding source code as string literals!

flake8_trio.py
jakkdl: `ast.withitem` doesn't have a `lineno` - so it's afaict not possible, changed it because it threw errors at me in tests.

Yeah the Bugbear way sounds good, will look into the boilerplate needed.

jakkdl: ah, the `context_expr` in the `withitem` has a lineno though.

jakkdl:
Okay, cribbed bugbear's testing setup. Had to replace the `namedtuple`+`partial` solution to get it to pass pyright, though likely possible to get a cleaner solution.
Dropped stdin support for the `filename` parameter that bugbear had, was way too much of a hassle to get that covered and I'm not sure how big the need for it is.
I've never done fuzz tests before, but copy-pasted to my heart's content and they seem to do something.
Added back xdist since fuzzers started taking time.

tests/test_flake8_trio.py
```diff
@@ -0,0 +1,66 @@
...
+        errors = list(plugin.run())
+        expected = self.errors(
+            TRIO100(7, 8, "trio.fail_after"),
+            TRIO100(12, 8, "trio.fail_after"),
+            TRIO100(14, 8, "trio.move_on_after"),
+        )
+        self.assertEqual(errors, expected)
+
+
+class TestFuzz(unittest.TestCase):
```
Zac-HD, line 47: I'd skip this by default - as you note it's pretty slow, and I think running it by hand occasionally is plenty.

flake8_trio.py
```diff
@@ -34,29 +33,68 @@ def is_trio_call(node: ast.AST, *names: str) -> Optional[str]:
...
+        self.lineno = lineno
+        self.col = col
+        self.message = message
+        self.err_type = type(Plugin)
+
+    def values(self):
+        return (self.lineno, self.col, self.message, self.err_type)
+
+
+class TRIO100(Error):
```
Zac-HD, line 93: I don't think this is worth the overhead when we're just going to work with tuples anyway.

If the iteration I had was unsatisfying, how about we include `Plugin` in each tuple in `Visitor.problems`, so that we can just `yield from visitor.problems`?

flake8_trio.py
```diff
@@ -34,29 +33,68 @@ def is_trio_call(node: ast.AST, *names: str) -> Optional[str]:
...
+            self._tree = tree
+
+    def load_file(self, filename: str) -> ast.AST:
+        """Loads the file in a way that auto-detects source encoding.
 
-    def run(self) -> Generator[tuple[int, int, str, Type[Any]], None, None]:
+        Stolen from flake8_import_order because it's good.
+        """
+        lines = pycodestyle.readlines(filename)
+        return ast.parse("".join(lines))
```
Zac-HD, line 73:
```suggestion
        with tokenize.open(filename) as f:
            return ast.parse(f.read())
```

I'm not actually a fan of the "on error assume UTF-8" approach; it seems likely to cause syntax problems later and doesn't actually fix the encoding error.

tests/test_flake8_trio.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def test_tree(self):
+        plugin = Plugin(tree=ast.parse(""))
+        errors = list(plugin.run())
+        self.assertEqual(errors, [])
+
+    def test_trio100(self):
+        filename = Path(__file__).absolute().parent / "trio100.py"
+        plugin = Plugin(filename=str(filename))
+        errors = list(plugin.run())
```
Zac-HD, line 27: Let's wrap this up in a helper `def get_errors_for_filename(fname: str) -> ast.AST:`, since we'll be doing it a lot.

Or even `def assert_expected_errors(self, fname: str, *expected: Tuple[...]) -> None:` to cover *all* the boilerplate.

tests/test_flake8_trio.py
jakkdl: Adding a `pytest.mark.fuzz` to them, so could maybe add `--runfuzz` to the release test (which I think only you can add?).

flake8_trio.py
jakkdl: It was because I was hitting typing problems with the `partial` solution that bugbear had, but yeah I simplified it down a lot and am just doing a tuple + factory function.

Lots of minor variations this could take, could do a `namedtuple` (but atm never accesses any members anyway so w/e), or could do away with the function and let callers do the format + run add the type.

flake8_trio.py
jakkdl: This one I more or less just straight copy-pasted w/o much investigation. Sounds good, and yeah it looked quite ugly after I skipped out on `stdin` support.

tests/test_flake8_trio.py
jakkdl: yeah was definitely gonna do that sooner or later, but I thought I might wait until I have a few more tests implemented to see how much of the work is identical between them and the best way to lay out the helper function. But I can push in an initial version now

jakkdl:
fixed all comments, code starting to look pretty :sparkle:

Zac-HD (approved):
Lovely!  Merging this now since it's a huge improvement on the status quo (😅), and I'll push a small refactoring commit in lieu of nitpicky review.

PR #4, opened 2022-07-18
Trio101: Never yield inside a nursery or cancel scope

jakkdl:
Not entirely sure dumping a bunch of yields into a set is the best way to avoid false positives, but other than doing the reverse (saving `contextmanagers`/`functiondefs`, or their linenos) I couldn't figure out a neater way of doing it.

Didn't bother trying to juggle out trio100 changes, so you might want to merge the first pull request before looking at this diff.

### DIFF
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -18,6 +18,13 @@ __version__ = "22.7.1"
 
 
 Error = Tuple[int, int, str, Type[Any]]
+cancel_scope_names = (
+    "fail_after",
+    "fail_at",
+    "move_on_after",
+    "move_at",
+    "CancelScope",
+)
 
 
 def make_error(error: str, lineno: int, col: int, *args: Any, **kwargs: Any) -> Error:
@@ -41,62 +48,60 @@ class Visitor(ast.NodeVisitor):
         super().__init__()
         self.problems: List[Error] = []
         self.safe_yields: Set[ast.Yield] = set()
+        self._yield_is_error = False
+        self._context_manager = False
+
+    def visit_generic_with(self, node: Union[ast.With, ast.AsyncWith]):
+        self.check_for_trio100(node)
+
+        outer = self._yield_is_error
+        if not self._context_manager and any(
+            is_trio_call(item, "open_nursery", *cancel_scope_names)
+            for item in (i.context_expr for i in node.items)
+        ):
+            self._yield_is_error = True
+
+        self.generic_visit(node)
+        self._yield_is_error = outer
 
     def visit_With(self, node: ast.With) -> None:
-        self.check_for_trio100(node)
-        self.check_for_trio101(node)
-        self.generic_visit(node)
+        self.visit_generic_with(node)
 
     def visit_AsyncWith(self, node: ast.AsyncWith) -> None:
-        self.check_for_trio100(node)
-        self.check_for_trio101(node)
-        self.generic_visit(node)
+        self.visit_generic_with(node)
 
-    def visit_FunctionDef(self, node: ast.FunctionDef) -> None:
-        self.trio101_mark_yields_safe(node)
-        self.generic_visit(node)
-
-    def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef) -> None:
-        self.trio101_mark_yields_safe(node)
-        self.generic_visit(node)
-
-    def check_for_trio100(self, node: Union[ast.With, ast.AsyncWith]) -> None:
-        # Context manager with no `await` call within
-        for item in (i.context_expr for i in node.items):
-            call = is_trio_call(item, "fail_after", "move_on_after")
-            if call and not any(isinstance(x, ast.Await) for x in ast.walk(node)):
-                self.problems.append(
-                    make_error(TRIO100, item.lineno, item.col_offset, call)
-                )
-
-    def trio101_mark_yields_safe(
+    def visit_generic_FunctionDef(
         self, node: Union[ast.FunctionDef, ast.AsyncFunctionDef]
-    ) -> None:
+    ):
+        outer = self._context_manager
         if any(
             isinstance(d, ast.Name)
             and d.id in ("contextmanager", "asynccontextmanager")
             for d in node.decorator_list
         ):
-            self.safe_yields.update(
-                {x for x in ast.walk(node) if isinstance(x, ast.Yield)}
-            )
+            self._context_manager = True
+        self.generic_visit(node)
+        self._context_manager = outer
 
-    def check_for_trio101(self, node: Union[ast.With, ast.AsyncWith]) -> None:
+    def visit_FunctionDef(self, node: ast.FunctionDef) -> None:
+        self.visit_generic_FunctionDef(node)
+
+    def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef) -> None:
+        self.visit_generic_FunctionDef(node)
+
+    def visit_Yield(self, node: ast.Yield) -> None:
+        if self._yield_is_error:
+            self.problems.append(make_error(TRIO101, node.lineno, node.col_offset))
+
+        self.generic_visit(node)
+
+    def check_for_trio100(self, node: Union[ast.With, ast.AsyncWith]) -> None:
+        # Context manager with no `await` call within
         for item in (i.context_expr for i in node.items):
-            call = is_trio_call(
-                item,
-                "open_nursery",
-                "fail_after",
-                "fail_at",
-                "move_on_after",
-                "move_at",
-            )
-            if call and any(
-                isinstance(x, ast.Yield) and x not in self.safe_yields
-                for x in ast.walk(node)
-            ):
+            call = is_trio_call(item, *cancel_scope_names)
+            if call and not any(isinstance(x, ast.Await) for x in ast.walk(node)):
                 self.problems.append(
-                    make_error(TRIO101, item.lineno, item.col_offset, call)
+                    make_error(TRIO100, item.lineno, item.col_offset, call)
                 )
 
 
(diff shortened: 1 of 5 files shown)


flake8_trio.py
```diff
@@ -34,29 +33,116 @@ def is_trio_call(node: ast.AST, *names: str) -> Optional[str]:
...
         self.generic_visit(node)
 
+    def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef) -> None:
+        self.trio101_mark_yields_safe(node)
+        self.generic_visit(node)
+
+    def check_for_trio100(self, node: Union[ast.With, ast.AsyncWith]) -> None:
+        # Context manager with no `await` call within
+        for item in (i.context_expr for i in node.items):
+            call = is_trio_call(item, "fail_after", "move_on_after")
```
Zac-HD, line 60: [Looks like we should also check for `CancelScope`, `fail_at`, and `move_on_at`](https://trio.readthedocs.io/en/stable/reference-core.html#cancellation-api-details).

Zac-HD (commented):
Will review properly after the other PR is merged, but it looks like a good approach 👍

flake8_trio.py
```diff
@@ -58,6 +69,36 @@ def check_for_trio100(self, node: Union[ast.With, ast.AsyncWith]) -> None:
...
+                item,
+                "open_nursery",
+                "fail_after",
+                "fail_at",
+                "move_on_after",
+                "move_at",
+            )
+            if call and any(
+                isinstance(x, ast.Yield) and x not in self.safe_yields
+                for x in ast.walk(node)
```
Zac-HD, line 96: Instead of collecting and marking all the yields, how about we track a `self._yield_is_error` boolean?  Then:

```python
    def visit_AsyncWith(self, node: ast.AsyncWith) -> None:
        self.check_for_trio100(node)
        outer = self._yield_is_error
        self._yield_is_error = any(is_trio_call(i.context_expr, "open_nursery") for i in node.items)
        self.generic_visit(node)
        self._yield_is_error = outer

    # similarly in visit_With for the sync context managers

    # visit_(Async)FunctionDef can set it to False, do generic_visit, set back to outer

    def visit_Yield(self, node: ast.Yield) -> None:
        if self._yield_is_error:
            self.problems.append(...)  # we want to point this error at the yield statement, not the cxt manager
        self.generic_visit(node)
```

Should be a bit cheaper since we avoid the extra iteration, and it's a generally useful pattern which I'd expect to come up again.

flake8_trio.py
```diff
@@ -79,3 +120,4 @@ def run(self) -> Generator[Tuple[int, int, str, Type[Any]], None, None]:
 
 
 TRIO100 = "TRIO100: {} context contains no checkpoints, add `await trio.sleep(0)`"
+TRIO101 = "TRIO101: {} never yield inside a nursery or cancel scope"
```
Zac-HD, line 123:
```suggestion
TRIO101 = "TRIO101: yield inside a {} context is only safe when implementing a context manager - otherwise, it breaks exception handling"
```

Zac-HD (commented):
- Requesting a larger design change sorry, but I think it's worth it.
- It'd also be good to avoid false alarms on `@(async)contextmanager`-decorated functions, in which cases yielding inside these contexts *is* OK after all.  Plus tests for that case.
- New error codes should be added to the README, with as long a message as is needed to explain what's going on and how to fix it.  (for that matter the existing TRIO100 message could use a link or two)

flake8_trio.py
jakkdl: hmm, `CancelScope` looks tricky to check for lexicographically - would have to parse out the value of everything that's `with`'d to see if it's a `CancelScope` defined somewhere else. But I can give it a try if you want

Zac-HD: Nah, I meant literally just checking for inline calls to CancelScope - `with trio.CancelScope() as scope:` is code that someone might write, and it's easy to check for *that* case.

jakkdl: ah right, makes sense

jakkdl:
> New error codes should be added to the README, with as long a message as is needed to explain what's going on and how to fix it. (for that matter the existing TRIO100 message could use a link or two)
I can give it a try, but not actually knowing Trio (and it usually not being too relevant for implementing the AST crawling) it'll require me to do a bunch of extra reading to write something educational (and correct). But I can give that a go later on.

Zac-HD:
> I can give it a try, but not actually knowing Trio (and it usually not being too relevant for implementing the AST crawling) it'll require me to do a bunch of extra reading to write something educational (and correct). But I can give that a go later on.

Let's just copy the error message from the code to the README then, and I'll improve the docs when we publish.

jakkdl:
Fixed. Small Q's:
Want to bother with mentioning inside what scope it is in the error as before? Could replace `self._yield_is_error` with a `str` with the scope name.

Unrelated to trio101 in particular, but:
Is TRIO100 really only for `fail_after` and `move_on_after`? I extended it to all cancel scopes except nursery for now.

The fuzzer is way slower after your rewrite, idk if it's checking more files or something but I killed it after it ran for 3 minutes w/o completion on my system.

Zac-HD:
- Ooh, yeah, mentioning which scope specifically would be lovely - ideally also with the line number (e.g. "`trio.move_on_after on line 5`")
- Extending to all cancel scopes is great 👍 
- I made the fuzzer run 100x more examples by default, feel free to turn that back down if it's excessive 😅

flake8_trio.py
```diff
@@ -40,19 +47,58 @@ class Visitor(ast.NodeVisitor):
...
+        self, node: Union[ast.FunctionDef, ast.AsyncFunctionDef]
+    ):
+        outer = self._context_manager
+        if any(
+            isinstance(d, ast.Name)
+            and d.id in ("contextmanager", "asynccontextmanager")
+            for d in node.decorator_list
+        ):
+            self._context_manager = True
+        self.generic_visit(node)
```
Zac-HD, line 83: We *also* want to un-set `self._yield_is_error` here; it's OK if you *define* a generator function inside a nursery or cancel scope - just not yield from it.

- If you're inside a contextmanager-decorated function, yield is OK
- Otherwise, if you're in a cancel scope or nursery, yield is an error
- *Unless* you're inside a function being defined in that scope, then it's OK again

jakkdl: isn't that the current behaviour? The default value is false, and it's only set to true within the cancel scope (if it's not w/in a context manager)

Zac-HD:
```python
async def f():
        yield 1  # is and should be an error

        def g():
            yield 2  # is but should not be an error
```

jakkdl: yep, you're right. Fixed~

jakkdl:
> * Ooh, yeah, mentioning which scope specifically would be lovely - ideally also with the line number (e.g. "`trio.move_on_after on line 5`")
Is there any official way to get markers on different lines in flake8? One of my regular linters (idk if it's pylint or mypy though lol) can put an error marker on one line that refers to another line that gets an info marker. Or we could add two "problems" one pointing to the yield and one pointing to the scope.
>     * Extending to all cancel scopes is great +1
And `open_nursery` with no await is fine?
>     * I made the fuzzer run 100x more examples by default, feel free to turn that back down if it's excessive sweat_smile
Hmm, I suppose that could be added as a parameter. But yeah lowering it to 1k (10x) tests still took my machine a minute :innocent:

jakkdl:
Any opinion on `self._yield_is_error`, `self._context_manager` and similar being a stack (list) and `push`/`pop`'ing values instead of storing them as "outers" in local variable in functions?

jakkdl:
Do you want
```
from trio import open_nursery
def foo9():
    with open_nursery:
        yield 1


import trio as mytrio
def foo10():
    with mytrio.open_nursery:
        yield 1
```
to be detected?

flake8_trio.py
```diff
@@ -40,19 +47,58 @@ class Visitor(ast.NodeVisitor):
...
+        self._yield_is_error = False
+        self._context_manager = False
 
-    def visit_With(self, node: ast.With) -> None:
+    def visit_generic_with(self, node: Union[ast.With, ast.AsyncWith]):
         self.check_for_trio100(node)
+
+        outer = self._yield_is_error
+        if not self._context_manager and any(
+            is_trio_call(item, "open_nursery", *cancel_scope_names)
```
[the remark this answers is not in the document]
Zac-HD: 🤷‍♂️ nah, doesn't seem worth the trouble when it'll crash at runtime.  For `flake8-trio` I'm mostly concerned with silent errors - the nightmare where your code runs, but (sometimes) doesn't do what you meant or expected.

Zac-HD:
> > Ooh, yeah, mentioning which scope specifically would be lovely - ideally also with the line number (e.g. "`trio.move_on_after on line 5`")
> 
> Is there any official way to get markers on different lines in flake8? One of my regular linters (idk if it's pylint or mypy though lol) can put an error marker on one line that refers to another line that gets an info marker. Or we could add two "problems" one pointing to the yield and one pointing to the scope.

I meant to have the marker point at the `yield`, and just mention the other location in the message like `"TRIO101: yield inside {call} (line {item.lineno}) is only safe ...`.

But let's revisit this later, it can be retrofitted in a later PR.

> > Extending to all cancel scopes is great +1
> 
> And `open_nursery` with no await is fine?

Yep, you might just want to e.g. call `nursery.start_soon(...)` a few times and that's a sync method.

Zac-HD (approved):
Looks great!  Merging now 😁

jakkdl:
> I meant to have the marker point at the yield, and just mention the other location in the message like "TRIO101: yield inside {call} (line {item.lineno}) is only safe ....

Ye I followed, but thought one could go a small step further and save people from having to look up line numbers - but unless there's a neat built-in way of doing it it might be ugly to raise two errors for one problem.

flake8_trio.py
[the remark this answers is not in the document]
jakkdl: Yeah that's what I figured, either do it all the way or ignore the distinction.

Issue #1, opened 2022-07-17
Ideas for `flake8-trio` checks


Zac-HD:
As a "list of ideas" issue, any of the ideas below might turn out to be too difficult to implement, or in fact counterproductive if implemented.  Nonetheless, we need somewhere to dump ideas and this issue is that place.  Items are in rough priority order by a combination of estimated importance and tractability.

- [x] [Never `yield` inside a nursery or cancel scope](https://github.com/python-trio/trio/issues/671#issuecomment-431646729)
  #4 
- [x] Anything that `await`s inside a `finally:` clause needs shielding *and* a timeout, e.g. `with trio.fail_after(1, shield=True):`  ([source](https://github.com/python-trio/trio/issues/671#issuecomment-578984615)) (also applies to `async for` or `async with`)
  #5
- [x] Never have an `except Cancelled` or `except BaseException` block with a code path that doesn't re-`raise` the error.
  #8
- [x] Calling a Trio async function without immediately `await`ing it.  Per [the discussion here](https://github.com/python-trio/trio/issues/671#issue-360692109) typecheckers have some support for the (difficult!) general case, but it's easy and useful to make a list of all Trio functions and check for e.g. `trio.sleep(0)` without an `await`.  We *don't* need to check `async for` or `async with` because using the sync syntax is a noisy error at runtime, unlike the checkpoint warnings.
  #10
- [x] Importing things from the main `trio` namespace in a way that isn't just `import trio` (this is really just to make the linter easier to implement, but it's consistent with ~99% of existing usage anyway)
  #11
- [x] Async functions must have at least one [checkpoint (i.e. `await`, `async for`, or `async with`)](https://trio.readthedocs.io/en/stable/reference-core.html#checkpoints) on every code path, unless an exception is raised (the same rule that Trio internals follow)
  #16
- [x] Async *iterables* (i.e. with a `yield`) should have at least one checkpoint in each loop iteration, and an additional checkpoint after the last iteration (unless an exception is raised, again like the Trio internals)
  #17
- [x] Expanding `TRIO102` - you need a shielded cancel scope to `await` inside an `except Cancelled/BaseException` block, not *just* `finally:`
  #18
- [x] Async function definitions should not have a `timeout` parameter - encourage use of `trio.fail_after()` or `trio.move_on_after()` instead.  ([`httpx` does OK here](https://www.python-httpx.org/advanced/#fine-tuning-the-configuration), but most uses are IMO a design mistake)
  #20
[...]

PR #5, opened 2022-07-19
Trio102: await inside finally needs shielding and timeout

jakkdl:
A first draft of trio102. check_for_102 is kind of a mess, with having to inspect a lot of places to find cancel scopes, if they have a timeout specified, and if shielding is set. (It appears there was a shield parameter in the past that now is gone?). It might be a bit defensive and checking the parameter of the trio call is probably silly since they require one.

It also fails on fuzzing site code, since they do have `await`'s in `finally`'s and this doesn't care if they're related much to trio at all. Maybe add a global check that the file has to import or mention trio?

`trio102.py` has a fun case on line 39 (that will fail on python3.8 atm) that idk if it's worth caring about?

And it adds duplicates and is generally kinda ugly, but thought I'd push a first draft to see if I've understood the check correctly and so.

### DIFF
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -43,7 +43,7 @@ class TrioScope:
         self.packagename = packagename
         self.variable_name: Optional[str] = None
         self.shielded: bool = False
-        self.timeout: bool = False
+        self.has_timeout: bool = False
 
         if self.funcname == "CancelScope":
             for kw in node.keywords:
@@ -52,9 +52,9 @@ class TrioScope:
                     self.shielded = kw.value.value
                 # sets to True even if timeout is explicitly set to inf
                 if kw.arg == "deadline":
-                    self.timeout = True
+                    self.has_timeout = True
         else:
-            self.timeout = True
+            self.has_timeout = True
 
     def __str__(self):
         # Not supporting other ways of importing trio
@@ -89,7 +89,7 @@ class Visitor102(ast.NodeVisitor):
     def __init__(self) -> None:
         super().__init__()
         self.problems: List[Error] = []
-        self._inside_finally: Optional[ast.Try] = None
+        self._inside_finally: bool = False
         self._scopes: List[TrioScope] = []
         self._context_manager = False
 
@@ -169,7 +169,7 @@ class Visitor102(ast.NodeVisitor):
         outer_scopes = self._scopes
 
         self._scopes = []
-        self._inside_finally = node
+        self._inside_finally = True
 
         for item in node.finalbody:
             self.visit(item)
@@ -180,12 +180,12 @@ class Visitor102(ast.NodeVisitor):
     def check_for_trio102(self, node: Union[ast.Await, ast.AsyncFor, ast.AsyncWith]):
         # if we're inside a finally, and not inside a context_manager, and we're not
         # inside a scope that doesn't have both a timeout and shield
-        if self._inside_finally is not None and not self._context_manager:
-            for scope in self._scopes:
-                if scope.timeout and scope.shielded:
-                    break
-            else:
-                self.problems.append(make_error(TRIO102, node.lineno, node.col_offset))
+        if (
+            self._inside_finally
+            and not self._context_manager
+            and not any(scope.has_timeout and scope.shielded for scope in self._scopes)
+        ):
+            self.problems.append(make_error(TRIO102, node.lineno, node.col_offset))
 
 
 class Visitor(ast.NodeVisitor):
@@ -277,4 +277,4 @@ class Plugin:
 
 TRIO100 = "TRIO100: {} context contains no checkpoints, add `await trio.sleep(0)`"
 TRIO101 = "TRIO101: yield inside a nursery or cancel scope is only safe when implementing a context manager - otherwise, it breaks exception handling"
-TRIO102 = "TRIO102: await in finally without a cancel scope and shielding"
+TRIO102 = "TRIO102: it's unsafe to await inside `finally:` unless you use a shielded cancel scope with a timeout"
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -184,7 +184,7 @@ class Visitor(ast.NodeVisitor):
         # There's no visit_Finally, so we need to manually visit the Try fields.
         # It's important to do self.visit instead of self.generic_visit since
         # the nodes in the fields might be registered elsewhere in this class.
-        for item in *node.body, *node.handlers, *node.orelse:
+        for item in (*node.body, *node.handlers, *node.orelse):
             self.visit(item)
 
         outer = self._inside_finally
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -14,7 +14,7 @@ import tokenize
 from typing import Any, Generator, List, Optional, Set, Tuple, Type, Union
 
 # CalVer: YY.month.patch, e.g. first release of July 2022 == "22.7.1"
-__version__ = "22.7.1"
+__version__ = "22.7.2"
 
 
 Error = Tuple[int, int, str, Type[Any]]
@@ -146,7 +146,6 @@ class Visitor(ast.NodeVisitor):
         self.generic_visit(node)
 
     def visit_With(self, node: ast.With) -> None:
-        # print('poop')
         self.visit_generic_with(node)
 
     def visit_AsyncWith(self, node: ast.AsyncWith) -> None:
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -11,7 +11,7 @@ Pairs well with flake8-async and flake8-bugbear.
 
 import ast
 import tokenize
-from typing import Any, Generator, List, Optional, Tuple, Type, Union
+from typing import Any, Generator, List, Optional, Set, Tuple, Type, Union
 
 # CalVer: YY.month.patch, e.g. first release of July 2022 == "22.7.1"
 __version__ = "22.7.1"
@@ -49,6 +49,10 @@ class Visitor(ast.NodeVisitor):
         self.check_for_trio100(node)
         self.generic_visit(node)
 
+    def visit_Try(self, node: ast.Try) -> None:
+        self.check_for_trio102(node)
+        self.generic_visit(node)
+
     def check_for_trio100(self, node: Union[ast.With, ast.AsyncWith]) -> None:
         # Context manager with no `await` call within
         for item in (i.context_expr for i in node.items):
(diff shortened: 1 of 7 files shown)


jakkdl:
I'll merge in main to this when you've unbroken it :P
Upon seeing `trio.CancelScope` I'll have to modify this one a bit

Zac-HD:
Fixed main, and left you a review on the other PR 😁

jakkdl:
for the shield value, I'll parse constant bools - but do you want to assume True or False for anything else? E.g.
```
        with trio.CancelScope(deadline=30, shield=(1==0)):
            await foo()
        myvar = False
        with trio.open_nursery(10) as s:
            s.shield = myvar
```

Zac-HD:
> for the shield value, I'll parse constant bools - but do you want to assume True or False for anything else?

If it's inside a `finally:` block, then it must *always* be shielded, so I'd want an error on anything that isn't literally `True`.  No need for a variable when you should always have exactly the same value!

I'd even be happy to go a step further and say that should be the *first* statement of the fail/move_on blocks (or passed as an argument to CancelScope), and should not otherwise be assigned to at all.

jakkdl:
> > for the shield value, I'll parse constant bools - but do you want to assume True or False for anything else?
> 
> If it's inside a `finally:` block, then it must _always_ be shielded, so I'd want an error on anything that isn't literally `True`. No need for a variable when you should always have exactly the same value!
> 
> I'd even be happy to go a step further and say that should be the _first_ statement of the fail/move_on blocks (or passed as an argument to CancelScope), and should not otherwise be assigned to at all.

I think I'll pass on enforcing it to be the first statement unless you insist, I think current solution is pretty clean.

jakkdl:
Big rewrite, I suspect you might dislike the added class but I I did not see a clean simpler solution.
 All tests fails due to coverage ... which I don't understand and failed to alleviate - would love a tip on what's going on there.
 
 I'm not loving the four separate class variables for tracking state, and them working differently: `yield_is_error` has the logic in the with&def, while the other three are facts about the state with logic at the point of adding the problem. I think I prefer the latter way with added reuse of the variables between checks, and if you agree I might convert the former.
 
 Will add lineno reference to the finally (and probs fix trio101 at the same time). Could possibly give another reference to the scope if it e.g. has a timeout but no shield.

jakkdl:
Oh, is this fine?
```
finally:
  with trio.CancelScope(deadline=10, shield=True):
    with trio.fail_after(5):
      await foo()
```

flake8_trio.py
```diff
@@ -55,19 +83,73 @@ def __init__(self) -> None:
...
+        # if we're inside a finally, and not inside a context_manager, and we're either not in a scope, or in a scope that doesn't have both a timeout and shield
+        if (
+            self._inside_finally is not None
+            and not self._context_manager
+            and (
+                self._scope is None
+                or (
+                    self._scope is not None
+                    and not (self._scope.timeout and self._scope.shielded)
+                )
```
Zac-HD, line 148:
```suggestion
                or not (self._scope.timeout and self._scope.shielded)
```

flake8_trio.py
```diff
@@ -55,19 +83,73 @@ def __init__(self) -> None:
...
+            and isinstance(node.value, ast.Constant)
         ):
-            self._yield_is_error = True
+            self._scope.shielded = node.value.value
+        self.generic_visit(node)
 
+    def visit_Await(self, node: ast.Await) -> None:
+        # if we're inside a finally, and not inside a context_manager, and we're either not in a scope, or in a scope that doesn't have both a timeout and shield
+        if (
+            self._inside_finally is not None
```
Zac-HD, line 141: I've just fixed my lazy writeup in #1, so it's my bad, but this check should also apply to `async for` and `async with` - i.e. all the `checkpoint_node_types` where the scheduler can kick in.

flake8_trio.py
```diff
@@ -55,19 +83,73 @@ def __init__(self) -> None:
...
+            and node.targets[0].value.id == self._scope.variable_name
+            and node.targets[0].attr == "shield"
+            and isinstance(node.value, ast.Constant)
         ):
-            self._yield_is_error = True
+            self._scope.shielded = node.value.value
+        self.generic_visit(node)
 
+    def visit_Await(self, node: ast.Await) -> None:
+        # if we're inside a finally, and not inside a context_manager, and we're either not in a scope, or in a scope that doesn't have both a timeout and shield
```
Zac-HD, line 139: Wrap comment at ~88 chars?

flake8_trio.py
```diff
@@ -55,19 +83,73 @@ def __init__(self) -> None:
...
+        outer_scope = self._scope
+
+        # Check for a `with trio.<scope_creater>`
+        trio_scope = next(
+            (
+                get_trio_scope(item, "open_nursery", *cancel_scope_names)
+                for item in (i.context_expr for i in node.items)
+            ),
+            None,
+        )
```
Zac-HD, line 102: The coverage issue is an uncovered _branch_ `96 -> exit`.  When I see this it's usually that you defined a lambda which was never called, but there's no lambda here!

I'm pretty sure it's that we never exhaust the iterable, and therefore never use the `None` fallback value - because `get_trio_scope` returns None if it's not a trio scope.  Incidentally I think this means we'll only consider the first item and therefore fail to detect the cancel scope in e.g. `with open(""), trio.fail_at(0):`.

At that point it's probably easier to write this as an explicit for-loop with a `break` 😅

Zac-HD:
> Oh, is this fine?
> 
> ```
> finally:
>   with trio.CancelScope(deadline=10, shield=True):
>     with trio.fail_after(5):
>       await foo()
> ```

Yep, that's fine - the outer shield protects it from the Cancelled exception that (might have) triggered the `finally` clause, and there's a finite deadline 👍

Zac-HD (commented):
I think you're right that there's no clean and simple solution for this - and I think the solution you've got looks pretty good.  Some detail-level comments below but at a high level I like it.

For the yield_is_error thing, sure, convert it if you wish, but let's do that in a subsequent PR?  I've seen too many new features get delayed by scope-creep 😁

jakkdl:
This should error, right?
```python
with trio.CancelScope(deadline=10, shield=True):
  try:
    pass
  finally:
    await foo()
```

Zac-HD:
> This should error, right?
> 
> ```python
> with trio.CancelScope(deadline=10, shield=True):
>   try:
>     pass
>   finally:
>     await foo()
> ```

Correct, it should error, because if `pass` happened to raise `Cancelled` (😨) then `await foo()` would fail.

flake8_trio.py
jakkdl: oops, yep!

flake8_trio.py
jakkdl: I'm surprised black doesn't handle it, though looks like it's a long-standing issue https://github.com/psf/black/issues/1713

flake8_trio.py
jakkdl: Ooh right, of course. Very nice that the coverage caught that issue. Thanks!
And yeah, added a case for that which previous version of the code did not handle.

Preposterous! Do you really think
```python
trio_scope = None
for item in (i.context_expr for i in node.items):
    trio_scope = get_trio_scope(item, "open_nursery", *cancel_scope_names)
    if trio_scope is not None:
        break
```
is better than
```python                                                                  
trio_scope = next(
   filter(
       None,
       (
           get_trio_scope(item, "open_nursery", *cancel_scope_names)
           for item in (i.context_expr for i in node.items)
       ),
   ),
   None,
)
```
??? :grin:

jakkdl:
> > Oh, is this fine?
> > ```
> > finally:
> >   with trio.CancelScope(deadline=10, shield=True):
> >     with trio.fail_after(5):
> >       await foo()
> > ```
> 
> Yep, that's fine - the outer shield protects it from the Cancelled exception that (might have) triggered the `finally` clause, and there's a finite deadline +1

Hmmm, I think that means I need to make scope a stack (which is saved away upon entering finally) and when await'ing search through the stack to see if any of them have shield+timeout.

This check certainly isn't getting simpler!

jakkdl:
Fixed all the comments, and I broke out 102 on an impulse. As you said we can revisit that later.
I also got rid of `visit_generic_with` and `visit_generic_functiondef` in favor of `visit_AsyncWith` calling `visit_With`. felt cleaner /shrug


The `finally` doesn't have it's own `lineno`, do you want me to check the `lineno` of the first statement in the `finalbody` and subtract 1 ... or maybe just skip referring to the line number of the finally.

flake8_trio.py
```diff
@@ -36,66 +36,206 @@ def make_error(error: str, lineno: int, col: int, *args: Any, **kwargs: Any) ->
...
+                # sets to True even if timeout is explicitly set to inf
+                if kw.arg == "deadline":
+                    self.timeout = True
+        else:
+            self.timeout = True
+
+    def __str__(self):
+        # Not supporting other ways of importing trio
+        # if self.packagename is None:
+        # return self.funcname
```
Zac-HD, line 62: This is reasonable, but we'll also want to lint on "some way of importing stuff from trio that doesn't have `trio.` at the start of each name".  Not a matter for this PR, but leave a comment on the issue?

flake8_trio.py
```diff
@@ -36,66 +36,206 @@ def make_error(error: str, lineno: int, col: int, *args: Any, **kwargs: Any) ->
...
 
-def is_trio_call(node: ast.AST, *names: str) -> Optional[str]:
+class TrioScope:
+    def __init__(self, node: ast.Call, funcname: str, packagename: str):
+        self.node = node
+        self.funcname = funcname
+        self.packagename = packagename
+        self.variable_name: Optional[str] = None
+        self.shielded: bool = False
+        self.timeout: bool = False
```
Zac-HD, line 46: Let's name this `has_timeout`, to avoid confusing it for the length of a timeout.

flake8_trio.py
```diff
@@ -129,10 +269,12 @@ def from_filename(cls, filename: str) -> "Plugin":
...
-        yield from visitor.problems
+        for v in (Visitor, Visitor102):
+            visitor = v()
+            visitor.visit(self._tree)
+            yield from visitor.problems
 
 
 TRIO100 = "TRIO100: {} context contains no checkpoints, add `await trio.sleep(0)`"
 TRIO101 = "TRIO101: yield inside a nursery or cancel scope is only safe when implementing a context manager - otherwise, it breaks exception handling"
+TRIO102 = "TRIO102: await in finally without a cancel scope and shielding"
```
Zac-HD, line 280:
```suggestion
TRIO102 = "TRIO102: it's unsafe to await inside `finally:` unless you use a shielded cancel scope with a timeout"
```

Zac-HD:
> The `finally` doesn't have it's own `lineno`, do you want me to check the `lineno` of the first statement in the `finalbody` and subtract 1 ... or maybe just skip referring to the line number of the finally.

Let's just skip referring to the lineno of `finally:`, it should be pretty obvious from context.

Zac-HD (approved):
Looks great!  

Couple of copyedit-level nitpicks below, but I'm planning to fix those and release later this evening so you can move on to the next meaningful check 🎉

flake8_trio.py
jakkdl: hmm, I specifically remember you mentioning in one of these pull requests that you didn't care to support anything but `import trio` -> `trio.funcname`. I had a version of this that was on it's way of working similar to how we're doing with context managers, and wouldn't be hard to reinstate that - at the risk of false positives.

Zac-HD: I mean that we should error on "from trio import" or "import trio as"!

jakkdl: Aah, got it. That's an easy one :)

PR #7, opened 2022-07-26
generate test environments for 3.8, 3.9 and 3.10

jakkdl:
To save the pain of pushing code only to have it fail on github CI version tests I looked up how to set up tests against different environments directly in tox, and this works well. Added a default envlist so `tox` doesn't take forever.
With this you'll have to slightly change the CI calls (e.g. `-e py38-test` in the python3.8 test environment).

I noticed you said 3.7 -- 3.11 back in your e-mail though. Running the code against 3.11 takes forever and a library uses a deprecated feature that fails the test, so that seems not worth it. But wanted to double-check whether you want to test against 3.7 or not.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,9 @@ To run our test suite please use tox.
 # Formatting and Linting
 tox -e check
 # Test Running
-tox -e test
+tox -e py310-test
+# Run checks and all tests
+tox -e ALL
 ```
 
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,9 +18,9 @@ To run our test suite please use tox.
 # Formatting and Linting
 tox -e check
 # Test Running
-tox -e py310-test
-# Run checks and all tests
-tox -e ALL
+tox -e test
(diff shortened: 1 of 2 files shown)


Zac-HD:
Nice, this does seem helpful... even easier if we can keep a plain `test` env that just runs on whatever the current Python version is, which I think just requires `[testenv:test,{py38, py39, py310}-test]` 🙂.  I'd also use this in the default envlist for convenience.

I decided against Python 3.7 support after that email on the basis that it's annoying to support and I doubt anyone still using it will also want to use flake8-trio.

jakkdl:
I think rather than having a duplicate `py310-test` (when running `tox -e ALL`) by adding another `test` env we rename `py310-test` to just be `test`.
That syntax didn't work, but `[testenv:{py38-, py39-,}test]` apparently did :sweat_smile:

Makes sense, 3.8 is a pain enough without multi-line `with`'s!

Also added more suggested flags to CONTRIBUTING.md

Zac-HD:
> Makes sense, 3.8 is a pain enough without multi-line `with`'s!

They can be multi-line if you use backslashes, just not commas-and-parens!  
(yep.  fixing this was a leading motivation for the PEG parser)

PR #8, opened 2022-07-27
Trio103: except Cancelled/BaseException must be re-raised

jakkdl:
First draft, think I like the approach - although not totally sure how nested `try`'s should be handled. And I haven't looked up how `raise .. from` should be handled.

This should probably have two different error messages and numbers, one for "reached end of block without re-raising" and another for "raising different error or returning".


flake8_trio.py
```diff
@@ -255,6 +255,99 @@ def check_for_trio100(self, node: Union[ast.With, ast.AsyncWith]) -> None:
...
+        self.unraised = True
+        for n in node.orelse:
+            self.visit(n)
+        # does orelse always raise correctly
+        orelse_raised = not self.unraised
+
+        # if both paths always raises, unset unraised.
+        self.unraised = not (body_raised and orelse_raised)
+
+    # checking for/while in this context doesn't make *that* much sense
```
jakkdl, line 342: I find the theoretical case for this highly amusing:
```python
try:
  pass
except trio.Cancelled:
  while <...>:
    if <...>:
      break
    raise
  else:
    raise
```
is in theory totally acceptable - but it's just a very funky if statement.

... nowait nevermind me, brainfart. This is not how [while/for] / else works. Need some logic for break, or scream if somebody tries to do something stupid like this. Will fix

jakkdl (commented):
(testing review/threading features for commenting on stuff)

flake8_trio.py
jakkdl: Okay I think I'll just disregard all `raise`'s inside loops. I can't come up with any pattern where that is good code, and handling the bad code is a pain after all. But any return's in there need to be screamed about.

jakkdl: thanks for coming to my TED talk

tests/trio103.py
```diff
@@ -0,0 +1,90 @@
...
+        else:
+            raise e
+
+    try:
+        pass
+    except BaseException as e:
+        try:
+            raise e
+        except trio.Cancelled as f:
+            raise f  # ???
```
Zac-HD, line 41: This is weird but OK.

tests/trio103.py
```diff
@@ -0,0 +1,90 @@
...
+            pass
+        except:
+            pass
+        finally:
+            raise e
+
+    try:
+        pass
+    except BaseException as e:
+        raise  # In theory safe? But godawful ugly so treat as error anyway
```
Zac-HD, line 80: No, this is fine, it's actually my preferred style since it doesn't add a frame to the traceback.

Bare `raise` also works whether-or-not you've got an `as` clause on the except handler.

tests/trio103.py
```diff
@@ -0,0 +1,90 @@
...
+    except trio.Cancelled:  # error
+        while var:
+            raise
+
+        for i in var:
+            if var:
+                return  # error
+
+
+# TODO: how to handle raise from?
```
Zac-HD, line 90: `raise ... from <a cancelled or BaseException>` should be an error, cancellations have to stay cancellations for correct handling.  We should have a distinct error message for that though, and maybe even a distinct error code.

Zac-HD:
Approach looks good overall!

tests/trio103.py
jakkdl: Wait, you like bare raises even when you have an `as` clause on the except handler? Bare raises are fine in general
`tests/trio103.py:5`
```python
except (SyntaxError, ValueError, BaseException):
    raise
```
But the case on line 80 is specifically for when you save it in a variable but do a bare raise anyway - in which case you should just skip the `as` clause entirely?
I could create a separate error message for it (~"don't do bare raise when you've saved an exception with an `as` clause"), but a good linter should pick up on the variable `e` being unused.

Zac-HD: Currently a super-niche case, but see https://peps.python.org/pep-0678/#example-usage

If there really is an unused `as` clause, it should get picked up by a _different_ flake8 plugin, since it's not a trio-related problem 🙂

tests/trio103.py
jakkdl: Thanks, fixed!

tests/trio103.py
jakkdl: Hmm, I guess that makes sense. I would've written `raise e` in that example, but I guess you don't have to and the bare `raise` is .. shorter. Feels like something a strongly opinionated formatting linter should pick up on.

Also further solidifies the case for not thoroughly parsing nested excepts.

But it's fixed! :+1:

tests/trio103.py
jakkdl: I kinda wanna print that out and put on the wall, I get that feeling so much trying to come up with corner cases - though a lot of them only look weird because the test cases are so stripped down. Somewhere out there there's a good case for writing deeply nested `except`'s

Quoted line 41 won't be considered an error, but I don't consider the parent exception covered.

jakkdl:
I think it's fully finished now, so review away!

I got almost all the way to a fully thorough parsing of nested `except`'s - but I stopped when I realized I needed to handle this case:
```python
except BaseException as e:
  try:
    raise
  except ValueError:
    raise # error, but parent except is fully covered if replaced with `raise e`
  except:
    raise e
```
But I can probably fix it if you want. Would require the beautiful `if` test in `visit_Raise` to get even bigger though :grin:

flake8_trio.py
```diff
@@ -255,6 +255,135 @@ def check_for_trio100(self, node: Union[ast.With, ast.AsyncWith]) -> None:
...
+                        and node.exc.func.id == "BaseException"
+                    )
+                    or (
+                        isinstance(node.exc.func, ast.Attribute)
+                        and isinstance(node.exc.func.value, ast.Name)
+                        and node.exc.func.value.id == "trio"
+                        and node.exc.func.attr == "Cancelled"
+                    )
+                )
+            )
```
Zac-HD, line 325: Pretty sure that it's never valid to raise a new exception!

For example this would allow replacing a `Cancelled` with a `BaseException`, which would definitely break things.

flake8_trio.py
```diff
@@ -255,6 +255,135 @@ def check_for_trio100(self, node: Union[ast.With, ast.AsyncWith]) -> None:
                 )
 
 
+# Never have an except Cancelled or except BaseException block with a code path that
+# doesn't re-raise the error
+class Visitor103_104(ast.NodeVisitor):
+    def __init__(self) -> None:
+        super().__init__()
+        self.problems: List[Error] = []
```
Zac-HD, line 263: Tempted to have a `Flake8TrioVisitor` superclass which sets up self.problems, and then we can iterate over it's `__subclasses__()` (iirc) in the Plugin.

flake8_trio.py
```diff
@@ -255,6 +255,135 @@ def check_for_trio100(self, node: Union[ast.With, ast.AsyncWith]) -> None:
...
+            self.visit(n)
+
+        # if body didn't raise, or it's unraised after else, set unraise
+        self.unraised = not body_raised or self.unraised
+
+    # disregard any raise's inside loops
+    def visit_For(self, node: ast.For):
+        outer_unraised = self.unraised
+        self.generic_visit(node)
+        self.unraised = outer_unraised
```
Zac-HD, line 379: Here (and in `visit_While`), we don't actually set `self.unraised`?

README.md
```diff
@@ -24,3 +24,5 @@ pip install flake8-trio
 - **TRIO101** `yield` inside a nursery or cancel scope is only safe when implementing a context manager - otherwise, it breaks exception handling.
 - **TRIO102** it's unsafe to await inside `finally:` unless you use a shielded
   cancel scope with a timeout"
+- **TRIO103** `except BaseException` and `except trio.Cancelled` with a code path that doesn't re-raise
```
Zac-HD, line 27: For the readme, where we can be more verbose, I think it's important to explain that we ignore any raises in loops: "Note  that any `raise` statement in a loop is ignored, as it's difficult to rule out the zero-iterations case".

flake8_trio.py
```diff
@@ -278,3 +407,5 @@ def run(self) -> Generator[Tuple[int, int, str, Type[Any]], None, None]:
 TRIO100 = "TRIO100: {} context contains no checkpoints, add `await trio.sleep(0)`"
 TRIO101 = "TRIO101: yield inside a nursery or cancel scope is only safe when implementing a context manager - otherwise, it breaks exception handling"
 TRIO102 = "TRIO102: it's unsafe to await inside `finally:` unless you use a shielded cancel scope with a timeout"
+TRIO103 = "TRIO103: except Cancelled or except BaseException block with a code path that doesn't re-raise the error"
+TRIO104 = "TRIO104: Cancelled and BaseException must be re-raised"
```
Zac-HD, line 411:
```suggestion
TRIO104 = "TRIO104: Cancelled (and therefore BaseException) must be re-raised"
```
I think the explanation does help to motivate this check.

--------

This does actually suggest that if you have an earlier `except:` clause for `Cancelled` we could do a more precise check.  However:

- I think that you should always re-raise BaseException *anyway* (e.g. for KeyboardInterrupt), and
- it's rare enough not to be worth the trouble here

tests/trio103_104.py
```diff
@@ -0,0 +1,102 @@
...
+except (SyntaxError, ValueError, BaseException):
+    raise
+except (SyntaxError, ValueError, trio.Cancelled) as p:  # error
+    pass
+except (SyntaxError, ValueError):
+    raise
+except trio.Cancelled as e:
+    raise e
+except:
+    pass
```
Zac-HD, line 14: Should be an error here, bare `except:` is equivalent to `except BaseException:` and we're not re-raising it!

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Changelog
 *[CalVer, YY.month.patch](https://calver.org/)*
 
+## 22.7.4
```
Zac-HD, line 4: If you've got multiple PRs open, might be easier to write this under a `## Future` heading and leave `__version__` alone, and I'll handle those after merging (and any merge conflicts).

[Note to self: also pull in [my usual ordering and consistent-increment checks](https://github.com/Zac-HD/hypothesis-jsonschema/blob/afbb2a919aceb8b784d532555946c6313e5be643/tests/test_version.py#L39-L55)]

flake8_trio.py
jakkdl: ooh right, yeah makes sense that would screw with trio - as would the reverse.
Though
```python
except trio.Cancelled:
  raise trio.Cancelled()
```
is fine in theory right? But if there's never/rarely any good reason to do that (instead of modifying the exception) can just simplify the logic.

flake8_trio.py
jakkdl: `__subclasses__()` works!

and could further clean up `Plugin.run` with a `classmethod`
```python
class TrioVisitor(ast.NodeVisitor):
    def __init__(self) -> None:
        super().__init__()
        self.problems: List[Error] = []

    @classmethod
    def run(cls, tree: ast.AST) -> Generator[Error, None, None]:
        visitor = cls()
        visitor.visit(tree)
        yield from visitor.problems
```
```python
def run(self) -> [...]:
    for v in TrioVisitor.__subclasses__():
        yield from v.run(self._tree)
```

flake8_trio.py
jakkdl:
```python
except trio.Cancelled:
  while [...]:
    if [...]:
      break
    raise
```
would set it and the `except` would think it's covered if it's not restored after the loop. Though I might have missed adding a test case for that

tests/trio103_104.py
jakkdl: oh, I thought bare `except:` was equivalent to `except Exception:` for some reason. Will fix!

CHANGELOG.md
jakkdl: Makes sense :+1:

flake8_trio.py
jakkdl: ooh that's a fun corner case. Don't think it would add too much to check for it though so I could add it

flake8_trio.py
Zac-HD: [Nope: per the Trio docs](https://trio.readthedocs.io/en/stable/reference-core.html#a-simple-timeout-example)

> every [Cancelled](https://trio.readthedocs.io/en/stable/reference-core.html#trio.Cancelled) object carries an invisible marker that makes sure that the cancel scope that triggered it is the only one that will catch it.

So you really *can't* replace the exception without breaking things.

flake8_trio.py
Zac-HD: Nah, not worth the code complexity IMO.

flake8_trio.py
Zac-HD: Code comment to this effect please, to help me out next time!

flake8_trio.py
Zac-HD: Nah, not worth the complexity (code + mental model) IMO, not when it's bad style anyway.

Zac-HD:
OK, I've merged the others and cherry-picked a few parts of this too; I'll let you merge or rebase as you prefer.

flake8_trio.py
jakkdl: Lol, gotta love invisible markers.

Fixed :+1:

README.md
jakkdl: It's actually `break` that's the biggest offender, the zero-iteration case (or `continue`s) can be covered if there's an `else` that always raises. So in theory it could accept loops that have no `break`s, an `else`, and both bodies always `raise` - but that's silly.

Fixed :+1:

flake8_trio.py
jakkdl: Yeah was a pretty terse comment - will try to comment more going forward!

Added a several comments to the other functions too ... catching a bug as I did :innocent:

flake8_trio.py
jakkdl: makes sense. Fixed :+1:

jakkdl:
Fixed all comments. Was sloppy and did that in the same commit as the merge, so hope that doesn't mess up your review habits too much.
I might wanna implement #13 before we merge this, the test file is a mess.

jakkdl:
Merged in #14 and was glad I didn't have any mismatches at first, but then I got to reorganizing and restructuring the tests and ooh boi that was some work. Also came up with a new way to avoid re-raising by abusing `continue`/`break` that I'll fix later.

tests/trio103_104.py
```diff
@@ -0,0 +1,174 @@
...
+    raise e
+# don't bypass raise by raising from nested except
+except trio.Cancelled as e:
+    try:
+        pass
+    except ValueError as g:
+        raise g  # error
+    except BaseException as h:
+        raise h  # error ?
+    raise e
```
jakkdl, line 95: Need help thinking on what is going on here. Is it always wrong to do `except <critical>` while inside a `except <critical>`?

tests/trio103_104.py
```diff
@@ -0,0 +1,174 @@
...
+try:
+    pass
+except BaseException:  # safe
+    while True:
+        break
+    raise
+except BaseException:  # safe
+    while True:
+        continue
+    raise
```
jakkdl, line 174: The new strange test cases I came up with, will likely have to track a loop stack to see that we don't break out from it.

jakkdl: okay that wasn't too bad

tests/trio103_104.py
Zac-HD: No, nesting like that can make sense depending on the contents of the inner `try:` block.

jakkdl: so: nested `raise` are not okay, unless the nested except is a handler for Cancelled/BaseException?

But hmm, this will bypass the Cancelled:
```python
try:
  ...
except trio.Cancelled:
  try:
    print(1/0)
  except:
    raise
```

Zac-HD: I propose we just ignore this for now - get the simple cases done in this PR, and open an issue for any follow-up which we can do (much) later.

jakkdl: sure - would just like an executive decision on whether to be silent about nested critical `except`s vs put up a warning and expect users to `#noqa`. Could add a third message "TRIO107 Warning: nested BaseException/trio.Cancelled" (or increment current 105&106 so they're adjacent - maybe good to not reserve error codes for WIP features in the future).
Putting that in would just be a few lines in `visit_ExceptHandler` so no big complexity hit.

If you just don't care / prefer status quo, you can just merge it as is - the PR should be done otherwise.

Zac-HD (approved):
This is probably going to complain about a *lot* of existing code... let's ship it!  Hands-on experience is usually the best way to work out if and how we might want to change a check 😁

jakkdl:
hehe, good luck! :grin:

PR #9, opened 2022-07-28
Merge trio102_py39.py into tests/trio102.py

jakkdl:
After your message https://github.com/Zac-HD/flake8-trio/issues/7#issuecomment-1197102185 I realized I'd completely misunderstood what the problem with `tests/trio100_py39.py` was on python3.8 - and that `tests/trio102_py39.py` as a separate file is completely superfluous if written without parentheses. So no reason to keep it around.


Zac-HD (approved):

Zac-HD:
Oops, yeah.  I feel that I should have spotted this in review too 😅

PR #10, opened 2022-07-28
TRIO105: Calling trio async function without await

jakkdl:
Would love any more problematic cases, or how you wanna handle assigning the `awaitable` to a variable. But the super basic case was indeed not that tricky~


README.md
```diff
@@ -24,3 +24,4 @@ pip install flake8-trio
 - **TRIO101** `yield` inside a nursery or cancel scope is only safe when implementing a context manager - otherwise, it breaks exception handling.
 - **TRIO102** it's unsafe to await inside `finally:` unless you use a shielded
   cancel scope with a timeout"
+- **TRIO105** Calling a trio async function without `await`ing it.
```
Zac-HD, line 27:
```suggestion
- **TRIO105** Calling a trio async function without immediately `await`ing it.
```

and the same in the code.

tests/trio105.py
```diff
@@ -0,0 +1,53 @@
...
+    # error
+    async with trio.open_file() as f:
+        pass
+
+    # currently errors out, but is unclear if user intended to save the
+    # awaitable or the result. Not too tricky to do a super basic parsing
+    # (save variable name, check if a variable with that name is ever awaited)
+    # but I suspect that isn't very useful regardless?
+    k = trio.open_file()
+    await k
```
Zac-HD, line 53: This should definitely be an error.

In cases where it's safe it's also entirely pointless - and very poor style to look dangerous + inconsistent.

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Changelog
 *[CalVer, YY.month.patch](https://calver.org/)*
 
+## 22.7.5
```
Zac-HD, line 4: [Note to self: do not merge this before releasing .4 or we'll end up with out-of-order releases 😱]

Zac-HD (approved):
Simple but effective 👍

jakkdl:
Fixed all :+1:

PR #11, opened 2022-07-28
Add 106, don't fancy import trio

jakkdl:
Uh, I'll just let you add this one to the changelog&readme, and increment version, with there being three simultaneous open pull requests.

### DIFF
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -14,7 +14,7 @@ import tokenize
 from typing import Any, Collection, Generator, List, Optional, Tuple, Type, Union
 
 # CalVer: YY.month.patch, e.g. first release of July 2022 == "22.7.1"
-__version__ = "22.7.3"
+__version__ = "22.7.4"
 
 
 Error = Tuple[int, int, str, Type[Any]]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 *[CalVer, YY.month.patch](https://calver.org/)*
 
-## Future
+## 22.7.4
 - Added TRIO105 check for not immediately `await`ing async trio functions.
 - Added TRIO106 check that trio is imported in a form that the plugin can easily parse.
 


tests/trio106.py
```diff
@@ -0,0 +1,13 @@
+import importlib
+
+import trio
+import trio as foo
+from trio import *  # noqa
+from trio import blah, open_file as foo  # noqa
```
Zac-HD, line 6: Ooh, this confused me for a minute.  Note to self: add a block comment below - "Note that our tests exercise the *Visitor* classes, without going through the noqa filter later in flake8 - so these suppressions are picked up by our project-wide linter stack but *not* the tests."

flake8_trio.py
```diff
@@ -278,3 +288,4 @@ def run(self) -> Generator[Tuple[int, int, str, Type[Any]], None, None]:
 TRIO100 = "TRIO100: {} context contains no checkpoints, add `await trio.sleep(0)`"
 TRIO101 = "TRIO101: yield inside a nursery or cancel scope is only safe when implementing a context manager - otherwise, it breaks exception handling"
 TRIO102 = "TRIO102: it's unsafe to await inside `finally:` unless you use a shielded cancel scope with a timeout"
+TRIO106 = "TRIO106: trio must be imported with `import trio` for the linter to work"
```
Zac-HD, line 291: Also needs a README entry

jakkdl:
While on it I resolved the above

tests/trio106.py
jakkdl: Yeah it confused me too - had to double-check that the flake8 filters got applied when ran as a plugin.

Zac-HD:
OK, apparently pushing commits marks the PR as closed rather than merged 😕 

It's merged to me though!

jakkdl:
Yeah that's some funky UI. It's marked as Closed, but it also says `Pull request successfully merged and closed`. But no matter

It looks like TRIO106 is missing from the readme though

jakkdl:
It could be worth adding a test that checks that each error message is mentioned in all four of the readme, changelog, plugin file and test file.

Zac-HD:
Sounds good, go for it!

jakkdl:
Opened an issue :+1:

PR #14, opened 2022-07-29
Meta-tests

### LINKED ISSUE #13: meta-tests
Check that all error messages (TRIO<XXX>) are mentioned in all four of `README.md` `CHANGELOG.md` `flake8_trio.py` and `tests/test_flake8_trio.py`.

Check that the line numbers of the set of lines in `tests/trio*.py` with an `# error` comment are the same as the line numbers in lines matching `make_error[...] <lineno> [...]` in `tests/test_flake8_trio.py`. As lines are moved around and specs change I get increasingly worried that these don't match.

jakkdl:
Fixes #13 

Also found what I think is a bug in TRIO100, and added TRIO106 to `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,8 @@ pip install flake8-trio
 - **TRIO100**: a `with trio.fail_after(...):` or `with trio.move_on_after(...):`
   context does not contain any `await` statements.  This makes it pointless, as
   the timeout can only be triggered by a checkpoint.
-- **TRIO101** `yield` inside a nursery or cancel scope is only safe when implementing a context manager - otherwise, it breaks exception handling.
-- **TRIO102** it's unsafe to await inside `finally:` unless you use a shielded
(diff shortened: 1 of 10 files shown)


Zac-HD (approved):
Nice ✅ 

Quick style note, if you're comfortable with pytest-style test functions I'm just as happy to use those as unittest classes.

jakkdl:
I haven't used either extensively enough to have too much of an opinion - but I needed access to a `msg` parameter to give more info and the first place I looked that had it was unittest.

Even better would be to have multiple different tests generated for each file, instead of it quitting on the first file/test pair that fails, but didn't bother spending time on figuring out how to do that.

Zac-HD:
`assert condition, "message"` is the pytest idiom fyi 🙂

Agree that it's not worth extending further though!

jakkdl:
Thank you, good to know!

PR #15, opened 2022-07-30
don't register plugin when running flake8 checks

### LINKED ISSUE #12: tox flake8 checker runs this plugin when checking
I just had a fun instance where I'd accidentally left a `breakpoint()` in my code when running `tox -e check` ... which led to flake8 crashing when trying to lint my code - since it's apparently including this plugin in it's checks. I looked at flake8 documentation if there was a way to disable a registered plugin, but somehow didn't see one, and the act of installing the plugin when creating the tox environment registers it in flake8.
`--disable=TRIO` looks like it just suppresses printing the errors, not disable it from running.

Feels like there should be an easy way of fixing this. But if not I think we can pass a flag parsed in `setup.py` by setting `install_command` in `tox.ini`, or use `setenv` to set an environment variable that's similarly parsed in `setup.py`.

jakkdl:
Yay, first thing I tried worked to perfection!
Fixes #12

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python3
 
-import os
 from pathlib import Path
 
 import setuptools
@@ -18,13 +17,6 @@ with local_file("flake8_trio.py").open("r") as o:
     else:
         raise AssertionError("__version__ must be set")
 
-
-if "FLAKE8_TRIO_NO_REGISTER" in os.environ and os.environ["FLAKE8_TRIO_NO_REGISTER"]:
-    entry_points = {}
-else:
-    entry_points = {
-        "flake8.extension": ["TRIO = flake8_trio:Plugin"],
-    }
 
 setuptools.setup(
     name="flake8-trio",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -18,6 +19,13 @@ def local_file(name: str) -> Path:
         raise AssertionError("__version__ must be set")
 
 
+if "FLAKE8_TRIO_NO_REGISTER" in os.environ and os.environ["FLAKE8_TRIO_NO_REGISTER"]:
```
Zac-HD, line 22:
```suggestion
if os.getenv("FLAKE8_TRIO_NO_REGISTER"):
```

Defaults to None if not present

Zac-HD (commented):
I really don't like this approach, since it adds a new configuration option (at install-time!).  

If `flake8 --extend-ignore=TRIO` also works I'd strongly prefer that; otherwise if you really need it I'll be sad but merge this.

jakkdl:
Yeah --extend-ignore just silences the errors.

But after another scouring of the tox configuration another time I found the proper way of doing it!! Need to wipe the old `check` environment after updating `tox.ini` to get one without the plugin installed - but after that it works to perfection :tada:

PR #17, opened 2022-08-01
7: async iterables must checkpoint after yields

jakkdl:
I continue doing stuff in a bit of an overkill way, but handling continue/break seems kind of important on this one. I gotta rush to a board game evening, so there's some cleanup and missing tests left, and probably needs a bunch more comments - but you can look at overall approach and look at tests. 

Also fixing a bunch of stuff for 107/108 .. and some other rando stuff too. Messy.

Also, I should tweak code redundancy and stuff in the tests - started getting seriously irritating to have to care about all other codes so wrote a quick&dirty solution for filtering.
Not loving splitting up the tests, but the diff printer wasn't smart enough to handle small diffs in large lists - might change that / write my own diff printer. And also frustrating to shift a hundred error messages if adding a line at the top of the file (although thank god for vim's Ctrl-A & Ctrl-X) - sliiiightly tempted to generate the tests from comments in the files exclusively, but *that* is likely overkill.

### DIFF
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -14,7 +14,7 @@ import tokenize
 from typing import Any, Dict, Iterable, List, NamedTuple, Optional, Tuple, Type, Union
 
 # CalVer: YY.month.patch, e.g. first release of July 2022 == "22.7.1"
-__version__ = "22.8.1"
+__version__ = "22.8.2"
 
 
 class Statement(NamedTuple):
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -11,13 +11,14 @@ Pairs well with flake8-async and flake8-bugbear.
 
 import ast
 import tokenize
-from typing import Any, Dict, Generator, Iterable, List, Optional, Tuple, Type, Union
+from typing import Any, Dict, Iterable, List, Optional, Tuple, Type, Union
 
 # CalVer: YY.month.patch, e.g. first release of July 2022 == "22.7.1"
 __version__ = "22.7.6"
 
-
 Error = Tuple[int, int, str, Type[Any]]
+
+
 checkpoint_node_types = (ast.Await, ast.AsyncFor, ast.AsyncWith)
 cancel_scope_names = (
     "fail_after",
@@ -43,7 +44,7 @@ class Flake8TrioVisitor(ast.NodeVisitor):
         self.suppress_errors = False
 
     @classmethod
-    def run(cls, tree: ast.AST) -> Generator[Error, None, None]:
+    def run(cls, tree: ast.AST) -> Iterable[Error]:
         visitor = cls()
         visitor.visit(tree)
         yield from visitor.problems
@@ -74,6 +75,10 @@ class Flake8TrioVisitor(ast.NodeVisitor):
     def set_state(self, attrs: Dict[str, Any]):
         for attr, value in attrs.items():
             setattr(self, attr, value)
+
+    def walk(self, *body: ast.AST) -> Iterable[ast.AST]:
+        for b in body:
+            yield from ast.walk(b)
 
 
 class TrioScope:
@@ -609,9 +614,9 @@ class Visitor107_108(Flake8TrioVisitor):
     def visit_Try(self, node: ast.Try):
         # check worst case try exception
         body_always_checkpoint = self.always_checkpoint
-        for n in (b for body in node.body for b in ast.walk(body)):
-            if isinstance(n, ast.Yield):
-                body_always_checkpoint = ("yield", n.lineno)
+        for inner_node in self.walk(*node.body):
+            if isinstance(inner_node, ast.Yield):
+                body_always_checkpoint = ("yield", inner_node.lineno)
                 break
 
         # check try body
@@ -758,7 +763,7 @@ class Plugin:
             source = f.read()
         return cls(ast.parse(source))
 
-    def run(self) -> Generator[Tuple[int, int, str, Type[Any]], None, None]:
+    def run(self) -> Iterable[Error]:
         for v in Flake8TrioVisitor.__subclasses__():
             yield from v.run(self._tree)
 
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -14,7 +14,7 @@ import tokenize
 from typing import Any, Dict, Iterable, List, Optional, Tuple, Type, Union
 
 # CalVer: YY.month.patch, e.g. first release of July 2022 == "22.7.1"
-__version__ = "22.7.6"
+__version__ = "22.8.1"
 
 Error = Tuple[int, int, str, Type[Any]]
 
@@ -40,14 +40,14 @@ def make_error(error: str, lineno: int, col: int, *args: Any, **kwargs: Any) -> 
 class Flake8TrioVisitor(ast.NodeVisitor):
     def __init__(self):
         super().__init__()
-        self.problems: List[Error] = []
+        self._problems: List[Error] = []
         self.suppress_errors = False
 
     @classmethod
     def run(cls, tree: ast.AST) -> Iterable[Error]:
         visitor = cls()
         visitor.visit(tree)
-        yield from visitor.problems
+        yield from visitor._problems
 
     def visit_nodes(
         self, *nodes: Union[ast.AST, Iterable[ast.AST]], generic: bool = False
@@ -65,12 +65,12 @@ class Flake8TrioVisitor(ast.NodeVisitor):
 
     def error(self, error: str, lineno: int, col: int, *args: Any, **kwargs: Any):
         if not self.suppress_errors:
-            self.problems.append(make_error(error, lineno, col, *args, **kwargs))
+            self._problems.append(make_error(error, lineno, col, *args, **kwargs))
 
     def get_state(self, *attrs: str) -> Dict[str, Any]:
         if not attrs:
             attrs = tuple(self.__dict__.keys())
-        return {attr: getattr(self, attr) for attr in attrs if attr != "problems"}
+        return {attr: getattr(self, attr) for attr in attrs if attr != "_problems"}
 
     def set_state(self, attrs: Dict[str, Any]):
         for attr, value in attrs.items():
@@ -158,8 +158,9 @@ class VisitorMiscChecks(Flake8TrioVisitor):
         self.visit_With(node)
 
     def visit_FunctionDef(self, node: Union[ast.FunctionDef, ast.AsyncFunctionDef]):
-        outer = self.get_state("_safe_decorator", "_yield_is_error")
+        outer = self.get_state()
         self._yield_is_error = False
+        self._inside_loop = False
 
         # check for @<context_manager_name> and @<library>.<context_manager_name>
         if has_decorator(node.decorator_list, *context_manager_names):
@@ -170,11 +171,12 @@ class VisitorMiscChecks(Flake8TrioVisitor):
         self.set_state(outer)
 
     def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef):
+        self.check_109(node.args)
         self.visit_FunctionDef(node)
 
     def visit_Yield(self, node: ast.Yield):
         if self._yield_is_error:
-            self.problems.append(make_error(TRIO101, node.lineno, node.col_offset))
+            self.error(TRIO101, node.lineno, node.col_offset)
 
         self.generic_visit(node)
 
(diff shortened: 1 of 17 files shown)


jakkdl:
I'm confused by the lack of a requirement of a checkpoint before the first yield, isn't the point to enforce that a call to an async function will always checkpoint? But a function
```python
async def foo():
  while True:
    yield 1
    await trio.sleep(1)
```
isn't a checkpoint the first time it's called.

flake8_trio.py
```diff
@@ -603,3 +732,4 @@ def run(self) -> Generator[Tuple[int, int, str, Type[Any]], None, None]:
 TRIO106 = "TRIO106: trio must be imported with `import trio` for the linter to work"
 TRIO107 = "TRIO107: Async functions must have at least one checkpoint on every code path, unless an exception is raised"
 TRIO108 = "TRIO108: Early return from async function must have at least one checkpoint on every code path before it."
+TRIO109 = "TRIO109: Async iterables should have a checkpoint after each yield."
```
Zac-HD, line 735: "one or more checkpoints before each yield, and after the last yield"

tests/trio109.py
```diff
@@ -0,0 +1,150 @@
...
+    yield
+
+
+async def foo4():
+    yield 1
+    await (yield 1)  # error
+
+
+# async with
+async def foo_async_with():  # safe, checkpoint on both enter and exit of async with
```
Zac-HD, line 32: Nooooot *always* the case; Trio promises that at least one of enter/exit will be checkpoints, but it might not be both.

I think we should "handle" this by documenting it as a limitation of the analysis and moving on, at least until we're otherwise finished.

Zac-HD (commented):
Generally seems good to me!

> Also fixing a bunch of stuff for 107/108 .. and some other rando stuff too. Messy.

If you can split stuff out in separate PRs that don't seem too prone to merge conflicts, that would make review a bit easier.  Or if there *are* conflicts but we expect to have a day to rebase in between, I guess; that's more of a judgment call.

(nothing jumps out as "should be a second PR" from this one, fwiw, this is a general principle)

> And also frustrating to shift a hundred error messages if adding a line at the top of the file - sliiiightly tempted to generate the tests from comments in the files exclusively, but _that_ is likely overkill.

No kill like overkill!  I'd do it 😁 

----

> I'm confused by the lack of a requirement of a checkpoint before the first yield, isn't the point to enforce that a call to an async function will always checkpoint?

I think you're right, and that example should trigger a lint warning.  Quoting the Trio docs:

> If you call an async function provided by Trio (await <something in trio>), and it doesn’t raise an exception, then it always acts as a checkpoint. (If it does raise an exception, it might act as a checkpoint or might not.)
> This includes async iterators: If you write `async for ... in <a trio object>`, then there will be at least one checkpoint before each iteration of the loop and one checkpoint after the last iteration.

```python
# In each example, *all* of the `await`s are required (though async-for or async-with could substitute)
async def minimal_ok():
    await ...
    yield
    await ...

async def multiple_ok():
    await ...
    yield
    await ...
    yield
    await ...

async def loop_standard_ok():
    for _ in range(5):
        await ...
        yield
    await ...

async def loop_reverse_ok():
    await ...
    for _ in range(5):
        yield
        await ...

async def same_line_ok():
    yield await ...
    await ...

async def same_line_yield_from_ok():
    yield from (x async for x in ...)  # yield from is *only* OK when it's an async iterable
    await ...
```

I'm not proposing to check for *exactly* one checkpoint after the last iteration, pretty sure that's not actually the meaning (but see https://github.com/python-trio/trio/issues/2388).

jakkdl:
> > I'm confused by the lack of a requirement of a checkpoint before the first yield, isn't the point to enforce that a call to an async function will always checkpoint?

> I think you're right, and that example should trigger a lint warning. Quoting the Trio docs:

Ooh nice, that means I can entirely remove 109 and merge it with 107 - such that async iterables simply are a special case of 107. That should simplify this beast down a little bit~

flake8_trio.py
```diff
@@ -472,108 +474,235 @@ def visit_Call(self, node: ast.Call):
...
+            self.has_yield,
+            self.suppress_errors,
+        ) = state
 
     def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef):
-        outer = self.all_await
+        outer = self.get_state()
 
         # do not require checkpointing if overloading
         self.all_await = has_decorator(node.decorator_list, "overload")
```
jakkdl, line 506: What's the guarantees on `[async]contextmanager`s? Checkpoint on enter and exit, or at least one of enter and exit?

Zac-HD: At least one of enter and exit.

tests/trio109.py
jakkdl: Fully handled :+1:

jakkdl:
I made a complete mess in the git history, and almost lost several hours of work today trying to clean that up. But recovered everything and everything is now squashed up into a single commit rebased onto master.
Should be fully working, and relatively clean. But do poke comments where a function or test needs better comments - I am starting to see the case for a simpler check, but making 107/108 cover complicated cases is way too fun to pass up :innocent: 
the 107/108 test files are kind of massive, and might be somewhat messy after I merged the yield tests I wrote into them, but I hope I've cleaned them up enough it's possible to get a decent overview of them.

The 107 error message might want to be more descriptive, or split back into a 109 for iterables, to be more verbose about needing checkpoints before and after yields. But the large rewrite merging the logic was well worth it regardless, caught some silly stuff with it.
Can split 108 into a 110 too, or even have three error messages for `return w/o checkpoint` / `return after yield with no checkpoint inbetween` / `yield after yield with no checkpoint inbetween`.

jakkdl:
Ooh and mayhaps even good to give a line indicator to (one of) the offending yields. E.g. `yield with no checkpoint since yield on line {}` / `yield with no checkpoint since function definition on line {}` / `function exit with no checkpoint since yield on line {}`

jakkdl:
Added details to TRIO108, if you like it I'll give the same treatment to 107.

flake8_trio.py
```diff
@@ -498,108 +507,242 @@ def visit_Call(self, node: ast.Call):
...
     def visit_Try(self, node: ast.Try):
-        if self.all_await:
-            self.generic_visit(node)
-            return
+        # check worst case try exception
+        body_always_checkpoint = self.always_checkpoint
+        for n in (b for body in node.body for b in ast.walk(body)):
+            if isinstance(n, ast.Yield):
+                body_always_checkpoint = ("yield", n.lineno)
+                break
```
jakkdl, line 615: Being able to combine the loop into one iter is nice to be able to break out easily, but there must be some way of doing the iteration cleaner.

Zac-HD:
Woke up this morning pretty unwell, unfortunately, so I probably won't review this before next week.  Feel free to add new PRs, jump over to pytest-exceptiongroup, or take some time off as suits you 🙂

flake8_trio.py
jakkdl: I'm still trying to come up with a solution to it with some friends for the fun of it, and got one that only needs 2 `for` but doesn't shortcut the innermost loop.
```python
for inner_node in itertools.chain(*(itertools.chain(ast.walk(body)) for body in node.body)):
```
But I wrote a wrapper for `ast.walk` that gave an easy-to-read solution that also is efficient.

jakkdl:
Aww, hoping for a speedy recovery :heart:  
No pressure on getting back ASAP, I'll find stuff to do :slightly_smiling_face: 

After some sleep I got a good solution for my ugly loop, and thought a bit more about the error codes. I think it largely comes down to if you want more or fewer error codes - I could in theory cut it all down to a single error code, changing TRIO107 from  
`async [function/iterable] must have at least one checkpoint on every code path, unless an expection is raised`  
to
`[exit] from async [function] with no guaranteed checkpoint since [function definition] on line [<same line>]`  
and for iterables  
`[exit] from async [iterable] with no guaranteed checkpoint since [yield] on line [<line>]`  

so there's a single error code
1. `[return/yield/exit] from async [function/iterable] with no guaranteed checkpoint since [function definition/yield] on line [<line>]`

Or create up to five error codes:
1. `async function must have at least one checkpoint on every code path, unless an exception is raised`
1. `return from async function must have guaranteed checkpoints on all code paths before it`
1. `async iterable with no checkpoint when exhausting the iterable. Add a checkpoint after yield on line <line>`
1. `yield from async iterable with no guaranteed checkpoint since entering the function on line `<line>`
1. `yield with no guaranteed checkpoint since yield on line `<line>`

A reasonable compromise is perhaps to merge 1-2, and 3-5, such that there's one error code for async functions and one for async iterables - but this will depend on your workflow and stuff.
Or do something like define these five as 2XY, with X=0 for async functions and X=1 for async iterables, so there's a lot of flexibility in what error codes to ignore.

flake8_trio.py
jakkdl:
```python
for inner_node in itertools.chain.from_iterable(map(mywalk, node.body)):
```
is probably the best we came up with, but I suspect this is still less readable than the code currently in place, so I'll just leave it all as is.

jakkdl:
Another thing I thought about, when there's multiple possible un-checkpointed paths before a yield, I could make it point out all problems (either through multiple problems on the same statement, or with a message that extends to arbitrary size.
Example:
```
async def foo():
  if ...:
    await foo()
    yield
  yield # error: no checkpoint since function def, and also no checkpoint since yield
```
Currently it's semi-arbitrary which one it will complain about, usually the most recent one.

Though would need to avoid stuff like
```
yield
yield # error no checkpoint since yield on $line-1
yield # error no checkpoint since yield on $line-1, and $line-2
yield # error no checkpoing since yield on $line-1, and $line-2, and $line-3
```
Probably something like "if complaining about multiple statements, only point out statements no other problem has complained about"

jakkdl:
Since I've touched on various smaller things in this PR (get/set_state being the most prominent one) there will be merge conflicts & some code needs to be changed for consistency - so probably don't try to merge this one at the same time as any one of the others without giving me an opportunity to rebase/merge inbetween.

jakkdl:
Re multiple possible problematic yields, it might also be nice to try and inform about `await`s that aren't part of guaranteed checkpoints - if they're on the path between statements that require checkpoints between them:
```
async def foo():
  if ...:
    await bar() # warning: not guaranteed
  return # error, no guaranteed checkpoint since funcDef.
```
silly in this case, but with more complicated stuff (continue/break/etc) and potential bugs in the check or stuff that's intentionally skipped due to complexity, it might be good to get feedback on stuff that a user might think is guaranteed. And reduce the number of instances of a user raising an issue "why doesn't it think this massive function is checkpointed?? It's clearly guaranteed, duhh".

Probably a separate PR, would require modifying a couple different things.

Zac-HD:
I think that having different codes for async functions vs iterables makes sense, no real opinion on whether that means we should have two or five though.  Which I guess suggests two for simplicity, albeit with custom/helpful messages?

> Another thing I thought about, when there's multiple possible un-checkpointed paths before a yield, I could make it point out all problems (either through multiple problems on the same statement, or with a message that extends to arbitrary size.

I think having multiple warnings on the same statement makes sense, and if each mention the "since xxx" location that should help users understand which code paths are missing checkpoints.  Only pointing to the most recent `yield` on each code path should fix your example, I think; each would point to the previous line and that in turn suggests the obvious "add a checkpoint between each" fix.

> Re multiple possible problematic yields, it might also be nice to try and inform about `await`s that aren't part of guaranteed checkpoints - if they're on the path between statements that require checkpoints between them:

I think we can probably get away without this - let's leave it out for now and we can revisit if anyone actually does complain.

Zac-HD:
Actually, slightly different request: this continues to be a huge PR (obviously for a nontrivial feature!).  Once you've merge in main, let's focus on getting the minimum viable version of this merged; I'm happy to wait for a follow-up before releasing if you think that would leave it unfinished.  I just can't meaningfully review a 1.5k-line diff touching several different things 😅

jakkdl:
Sorry yeah I'll get it in working shape and split off into separate pull requests. Requires so much discipline to work this way!
Though the line count will stay kinda huge due to the test files being hundreds of lines long

jakkdl:
Down to 1000 lines added! :tada: :joy:

The main stuff is the +210/-58 in flake8_trio.py - tried commenting up stuff so it should be possible to follow the logic (and ofc caught a minor bug in the process).
The test files are sorta messy, having copied tests back and forth between 107/108/109 multiple times, but I did another cleanup pass and there shouldn't be too many redundant tests in them now. Also added a bunch of comments, I'm hopefully not too too deep into the weeds now and incapable of judging what checkpoints are unintuitive or not :sweat_smile: 

Don't think I can meaningfully split off more since 107&108 are so intertwined now :)

I think/hope that transitioning to multiple errors per line is gonna be quick, I'd already started on this branch originally (:flushed:). But branched that off from this one~

Zac-HD (approved):
Nice!  Big diff, but all the prep work has paid off 🙂

flake8_trio.py
```diff
@@ -561,105 +567,251 @@ def visit_Call(self, node: ast.Call):
...
-        self.all_await = outer
+    def check_function_exit(self, node: Union[ast.Return, ast.AsyncFunctionDef]):
+        # error if function exits w/o guaranteed checkpoint since function entry
+        method = "return" if isinstance(node, ast.Return) else "exit"
+
+        if self.always_checkpoint is not None:
+            if self.yield_count:
+                self.error("TRIO108", node, method, self.always_checkpoint)
+            else:
+                self.error("TRIO107", node, method, self.always_checkpoint)
```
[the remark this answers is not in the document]
jakkdl: I think I had that change in the multi-error branch ^^;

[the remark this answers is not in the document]
Zac-HD: Not a big deal anyway or I would have committed it before merging; more of a style note for myself.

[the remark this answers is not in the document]
jakkdl: yeah I'll fix with multi-error PR~

PR #18, opened 2022-08-02
Extend 102 to excepts, 104 also catch yield

jakkdl:
Also add info to error messages on 102&103, rename variables and minor refactor of Visitor102 to match later visitors.

### DIFF
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -14,7 +14,7 @@ import tokenize
 from typing import Any, Generator, Iterable, List, Optional, Tuple, Type, Union
 
 # CalVer: YY.month.patch, e.g. first release of July 2022 == "22.7.1"
-__version__ = "22.7.5"
+__version__ = "22.7.6"
 
 
 Error = Tuple[int, int, str, Type[Any]]
@@ -37,7 +37,7 @@ def make_error(error: str, lineno: int, col: int, *args: Any, **kwargs: Any) -> 
 
 
 class Flake8TrioVisitor(ast.NodeVisitor):
-    def __init__(self) -> None:
+    def __init__(self):
         super().__init__()
         self.problems: List[Error] = []
 
@@ -47,12 +47,19 @@ class Flake8TrioVisitor(ast.NodeVisitor):
         visitor.visit(tree)
         yield from visitor.problems
 
-    def visit_nodes(self, nodes: Union[ast.AST, Iterable[ast.AST]]) -> None:
-        if isinstance(nodes, ast.AST):
-            self.visit(nodes)
+    def visit_nodes(
+        self, *nodes: Union[ast.AST, Iterable[ast.AST]], generic: bool = False
+    ):
+        if generic:
+            visit = self.generic_visit
         else:
-            for node in nodes:
-                self.visit(node)
+            visit = self.visit
+        for arg in nodes:
+            if isinstance(arg, ast.AST):
+                visit(arg)
+            else:
+                for node in arg:
+                    visit(node)
 
     def error(self, error: str, lineno: int, col: int, *args: Any, **kwargs: Any):
         self.problems.append(make_error(error, lineno, col, *args, **kwargs))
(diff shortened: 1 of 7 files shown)


flake8_trio.py
```diff
@@ -37,7 +37,7 @@ def make_error(error: str, lineno: int, col: int, *args: Any, **kwargs: Any) ->
 
 
 class Flake8TrioVisitor(ast.NodeVisitor):
-    def __init__(self) -> None:
```
jakkdl, line 40: I was originally also running my code with `mypy --strict` - which is much stricter than pyright and also enforces explicit return values. But I transitioned to only running the code through pyright and especially usage of ` -> None` became inconsistent, so I thought it best to just clean up all the old instances of it

flake8_trio.py
```diff
@@ -185,112 +188,153 @@ def visit_Import(self, node: ast.Import):
...
 class Visitor102(Flake8TrioVisitor):
-    def __init__(self) -> None:
+    def __init__(self):
         super().__init__()
-        self._inside_finally: bool = False
-        self._scopes: List[TrioScope] = []
-        self._context_manager = False
+        self._critical_scope: Optional[Tuple[int, int, str]] = None
+        self._trio_context_managers: List[TrioScope] = []
+        self._safe_decorator = False
```
jakkdl, line 226: I thought through these names again, and realized that both decorators and `with` are context managers, so I think these names are better.

flake8_trio.py
```diff
@@ -185,112 +188,153 @@ def visit_Import(self, node: ast.Import):
...
 
-    def visit_Assign(self, node: ast.Assign) -> None:
```
jakkdl, line 195: moved down since it's not the key part of the class

flake8_trio.py
```diff
@@ -596,8 +622,8 @@ def run(self) -> Generator[Tuple[int, int, str, Type[Any]], None, None]:
 
 TRIO100 = "TRIO100: {} context contains no checkpoints, add `await trio.sleep(0)`"
 TRIO101 = "TRIO101: yield inside a nursery or cancel scope is only safe when implementing a context manager - otherwise, it breaks exception handling"
-TRIO102 = "TRIO102: it's unsafe to await inside `finally:` unless you use a shielded cancel scope with a timeout"
-TRIO103 = "TRIO103: except Cancelled or except BaseException block with a code path that doesn't re-raise the error"
+TRIO102 = "TRIO102: await inside {2} @ line,col {0},{1} must have shielded cancel scope with a timeout"
+TRIO103 = "TRIO103: {} block with a code path that doesn't re-raise the error"
 TRIO104 = "TRIO104: Cancelled (and therefore BaseException) must be re-raised"
 TRIO105 = "TRIO105: Trio async function {} must be immediately awaited"
 TRIO106 = "TRIO106: trio must be imported with `import trio` for the linter to work"
```
jakkdl, line 629: Finally got around to polishing up error messages a bit more, though unsure of exact format of `inside BaseException @ line,col 120, 4`. Column is likely overkill?
101 and 104 can get the same treatment, but didn't wanna update all the messages until settled.

Zac-HD: "inside X on line Y" seems good to me, column information doesn't seem worth the verbosity.

tests/test_flake8_trio.py
```diff
@@ -40,9 +40,19 @@ def trim_messages(messages: Iterable[Error]):
         errors = tuple(plugin.run())
 
         # start with a check with trimmed errors that will make for smaller diff messages
-        trim_errors = trim_messages(plugin.run())
+        trim_errors = trim_messages(errors)
         trim_expected = trim_messages(expected)
-        self.assertTupleEqual(trim_errors, trim_expected)
+
+        #
```
Zac-HD, line 46: missing comment

tests/trio102.py
```diff
@@ -122,3 +122,39 @@ async def foo3():
...
+
+
+async def foo5():
+    await foo()  # avoid TRIO107
+    try:
+        ...
+    except trio.Cancelled:
+        with trio.CancelScope(deadline=30, shield=True):
+            await foo()  # safe
+            raise  # avoid TRIO103
```
Zac-HD, line 152: dedent these for clarity?  the `raise` doesn't need to be inside the cancelscope.

Zac-HD (commented):
Two small suggestions below and one above; once those are handled it's ready to merge 🎉

tests/test_flake8_trio.py
jakkdl: This function is getting a makeover in #19 (which has merged this branch), so I'd leave it for the moment and avoid a merge conflict.

tests/trio102.py
jakkdl: Actually, lemme fix this in #19 too (I'm touching all those lines and getting rid of the `raise`s completely).

jakkdl: nvm no I'm not. 1 sec I'll fix

PR #20, opened 2022-08-05
Add trio[300], async functions should not have a timeout parameter

jakkdl:
Easy check, only question mark is if `*timeout` or `**timeout` is worthy of raising a problem.
Have not incremented version number, and 300 is a temporary error code until we've sorted out what to do with #17

### DIFF
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -157,6 +157,7 @@ class VisitorMiscChecks(Flake8TrioVisitor):
         self._safe_decorator, self._yield_is_error = outer
 
     def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef):
+        self.check_300(node.args)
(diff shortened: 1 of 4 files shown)


flake8_trio.py
```diff
@@ -187,6 +188,11 @@ def visit_Import(self, node: ast.Import):
             if name.name == "trio" and name.asname is not None:
                 self.problems.append(make_error(TRIO106, node.lineno, node.col_offset))
 
+    def check_300(self, args: ast.arguments):
+        for arg in (*args.posonlyargs, *args.args, *args.kwonlyargs):
```
Zac-HD, line 192: LGTM, a `*timeout` or `**timeout` param seems semantically different to me.

jakkdl:
fixed all :+1:

Zac-HD (approved):

PR #21, opened 2022-08-05
10: add trio[301], don't await trio.sleep inside loop

jakkdl:
Straightforward stuff. Made it also check [async] for, and to specifically only check "trio.sleep"


Zac-HD:
In order to stick to cases where Event really does seem better, I'd confine this check specifically to while-loops, and only where `await trio.sleep[_until](...)` is the *entire* loop body.

jakkdl:
Ah, fixed!

Zac-HD:
Nice!  Straightforward as you say, modulo some brain-fog-inspired complications merging several PRs in a row 😅

PR #25, opened 2022-08-06
various code changes/cleanup, update contributing

jakkdl:
I'm not in love with having to put quotation marks around strings in tests, but I think it's relatively clean that the text is straight up `eval`d.

There's probably a much better name out there for `HasLineInfo`, but I failed to come up with one.

Will likely update this one in a bit with something that works for multiple errors on the same line.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -25,6 +25,14 @@ tox -e ALL
 tox -e ALL -p --develop
 ```
 
+## Meta-tests
+To check that all codes are tested and documented there's a test that error codes mentioned in `README.md`, `CHANGELOG.md` (matching `TRIO\d\d\d`), the keys in `flake8_trio.Error_codes` and codes parsed from filenames matching `tests/trio*.py`, are all equal.
+
+## Test generator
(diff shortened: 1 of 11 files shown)


jakkdl:
Agh, black/shed + `trio*.py` files is frustrating. Shed doesn't let black read config file, and doesn't have it's own exclude flag (nor support regex/globs in file names), so only way I found to have shed not reformat a file was to explicitly list all other files - in which case it's probably better to ditch shed and run the underlying formatters on their own.

But I managed to change my error-generating line to a version that black wouldn't touch - so we're safe for now.

Zac-HD:
Like Black (which it's wrapping), Shed doesn’t reformat lines that end with `# fmt: skip` or blocks that start with `# fmt: off` and end with `# fmt: on`. Sorry for the frustration so far, but hopefully this will reduce it in future!

jakkdl:
Aha, thanks! Failed to find that one when looking. (though `# fmt: skip` didn't play well, but on/off worked)

Had a very fun one where it reformatted
```python
async with (yield):  # error: 16, "yield", Statement("function definition", lineno-3)
```
to
```python
async with (
    yield  # error: 16, "yield", Statement("function definition", lineno-3)
):
```
but that means the column number is wrong, so I change it to 8 ... upon which it now thinks it fits on one line and formats it back to the original .. which means the 8 should be a 16 again... :upside_down_face:

Zac-HD (approved):
Nice!

PR #27, opened 2022-08-08
More detailed error printing on failed tests

jakkdl:
Branched off of multi-problem PR, further improving error printing and catching some type errors here and there, although this is once again hitting overkill territory especially with not that many more tests left to create.

### DIFF
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -70,13 +70,7 @@ class TrioScope:
 
 
 class Error:
-    def __init__(
-        self,
-        error_code: str,
-        lineno: int,
-        col: int,
-        *args: Union[HasLineInfo, TrioScope, str, int],
-    ):
+    def __init__(self, error_code: str, lineno: int, col: int, *args: object):
         self.line = lineno
         self.col = col
         self.code = error_code
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -100,7 +100,7 @@ cancel_scope_names = (
     "fail_after",
     "fail_at",
     "move_on_after",
-    "move_at",
+    "move_on_at",
     "CancelScope",
 )
 context_manager_names = (
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -11,20 +11,10 @@ Pairs well with flake8-async and flake8-bugbear.
 
 import ast
 import tokenize
-from typing import Any, Dict, Iterable, List, NamedTuple, Optional, Tuple, Type, Union
+from typing import Any, Dict, Iterable, List, NamedTuple, Optional, Union
 
 # CalVer: YY.month.patch, e.g. first release of July 2022 == "22.7.1"
 __version__ = "22.8.2"
-
-
-class Statement(NamedTuple):
-    name: str
-    lineno: int
-    col_offset: int = 0
-
-
-HasLineInfo = Union[ast.expr, ast.stmt, ast.arg, ast.excepthandler, Statement]
-Error = Tuple[int, int, str, Type[Any]]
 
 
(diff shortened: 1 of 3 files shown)


flake8_trio.py
```diff
@@ -41,6 +31,79 @@ class Statement(NamedTuple):
...
+HasLineInfo = Union[ast.expr, ast.stmt, ast.arg, ast.excepthandler, Statement]
+
+
+class TrioScope:
+    def __init__(self, node: ast.Call, funcname: str):
+        self.node = node
+        self.funcname = funcname
+        self.variable_name: Optional[str] = None
+        self.shielded: bool = False
+        self.has_timeout: bool = False
```
Zac-HD, line 57: I'd default this to `True`, then set it to `False` inside the if-CancelScope block and drop the else.  Also add a comment that handling `scope.shield = True` is done elsewhere.

flake8_trio.py
```diff
@@ -41,6 +31,79 @@ class Statement(NamedTuple):
     "TRIO110": "`while <condition>: await trio.sleep()` should be replaced by a `trio.Event`.",
 }
 
+
+class Statement(NamedTuple):
+    name: str
+    lineno: int
+    col_offset: int = 0
+
+    # ignore col offset since many tests don't supply that
```
Zac-HD, line 40: `and (self.col_offset == other.col_offset or None in [self.col_offset, other.col_offset])`?

flake8_trio.py
```diff
@@ -41,6 +31,79 @@ class Statement(NamedTuple):
...
+                if kw.arg == "shield" and isinstance(kw.value, ast.Constant):
+                    self.shielded = kw.value.value
+                # sets to True even if timeout is explicitly set to inf
+                if kw.arg == "deadline":
+                    self.has_timeout = True
+        else:
+            self.has_timeout = True
+
+    def __str__(self):
+        # Not supporting other ways of importing trio
```
Zac-HD, line 71: ", see TRIO106"

flake8_trio.py
```diff
@@ -90,11 +144,14 @@ def visit_nodes(
                 for node in arg:
                     visit(node)
 
-    def error(self, error: str, node: HasLineInfo, *args: Any, **kwargs: Any):
+    def error(
+        self,
+        error: str,
+        node: HasLineInfo,
+        *args: Union[HasLineInfo, TrioScope, str, int],
```
Zac-HD, line 151: I'd usually just annotate that `Any`, or `object`, because it looks like a "whatever we get passed" argument and ultimately gets formatted into a string.

flake8_trio.py
```diff
@@ -592,12 +624,10 @@ def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef):
     def check_function_exit(self, node: Union[ast.Return, ast.AsyncFunctionDef]):
         # error if function exits w/o guaranteed checkpoint since function entry
         method = "return" if isinstance(node, ast.Return) else "exit"
+        error_code = "TRIO108" if self.yield_count else "TRIO107"
```
Zac-HD, line 627: Small perf optimization, but it's equally clear code: move this inside the if-statement, so we only compute it in the (relatively rare) cases where we actually use the results.

Zac-HD:
(just going to push a commit, merge, merge-master on #28)

flake8_trio.py
jakkdl: I had some reason for wanting to typecheck this when I wrote this, though it might've gotten lost in a refactor of `test_flake8_trio` and yeah it's pretty messy anyway.
If anything some kind of "is_stringable"

PR #28, opened 2022-08-08
7 async iterable checkpoints

jakkdl:
On top of #27 so the diff is reasonable large.
Needed to improve the printing a bit to get a handle on the errors flying around, but the extra logic wasn't too bad.

Also solved the case of lazily evaluated boolops


Zac-HD (approved):
Should make it much easier to understand what's going on here, if somewhat noiser too at first!

PR #30, opened 2022-08-09
pylint pass

jakkdl:
Looks like there's bugs in both mypy and pylint with regards to the `__eq__` in Statement in flake8_trio.py (nobody mentions NamedTuple in https://github.com/python/mypy/issues/6910 - but I think it's the same?), and while the pylint issue https://github.com/PyCQA/pylint/issues/225 is closed it's cropping up in `checkpoint_continue/break` as well.
But with 11 disabled flags flake8_trio and tests/test* now pass without any errors - though if pylint was added to tox a few of those would probably be `#pylint: disable=XXX` instead.

I elected to explicitly set encoding in all opens as well, according to https://peps.python.org/pep-0597/

I also wiped a redundant test after running `pylint --disable=all --enable=similarities tests/trio*` - but that also gives a bunch of false positives. Though I am tempted to merge 103&104, and instead of doing `#error:` to mark errors do `#TRIO\d\d\d:`. I think this would also allow more `#INCLUDE` (or rather, change it to `#EXCLUDE` and only use that when necessary) and that might catch more stuff that's missing an explicit check.

### DIFF
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -50,11 +50,17 @@ Error_codes = {
 class Statement(NamedTuple):
     name: str
     lineno: int
-    col_offset: int = 0
+    col_offset: int = -1
 
-    # ignore col offset since many tests don't supply that
     def __eq__(self, other: Any) -> bool:
-        return isinstance(other, Statement) and self[:2] == other[:2]
+        return (
+            isinstance(other, Statement)
+            and self[:2] == other[:2]
+            and (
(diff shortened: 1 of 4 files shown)


Zac-HD (approved):
This all looks reasonable to me, but let's not bother with further changes; it doesn't add much value and we're nearly done anyway.

jakkdl:
makes sense, I'll just run pylint locally then to catch any bugs while developing :+1:

PR #34, opened 2022-08-09
TRIO112: replace one-line nursery with regular function call

jakkdl:
Final remaining issue without a PR :muscle: 

Code is fugly, will make a pass tomorrow, but I think the check is done.

### DIFF
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -11,7 +11,18 @@ Pairs well with flake8-async and flake8-bugbear.
 
 import ast
 import tokenize
-from typing import Any, Dict, Iterable, List, NamedTuple, Optional, Set, Union
+from typing import (
+    Any,
+    Dict,
+    Iterable,
+    List,
+    NamedTuple,
+    Optional,
+    Set,
+    Tuple,
+    Union,
+    cast,
+)
 
 # CalVer: YY.month.patch, e.g. first release of July 2022 == "22.7.1"
 __version__ = "22.8.4"
@@ -44,7 +55,7 @@ Error_codes = {
         "`trio.[fail/move_on]_[after/at]` instead"
     ),
     "TRIO110": "`while <condition>: await trio.sleep()` should be replaced by a `trio.Event`.",
-    "TRIO112": "consider using a regular function call instead",
+    "TRIO112": "Redundant nursery {}, consider replacing with a regular function call",
 }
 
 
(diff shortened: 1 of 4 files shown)


tests/trio112.py
```diff
@@ -0,0 +1,35 @@
...
+
+with trio.open_nursery() as n:  # error: 0
+    n.start_soon(n=7)
+
+with trio.open_nursery() as n:  # error in theory
+    n.start_soon(lambda n: n + 1)
+
+# safe?
+with trio.open_nursery() as n, open("") as o:
+    n.start(...)
```
Zac-HD, line 17: Bit suspicious for #22-like reasons, but since violating this rule is more of a style problem than a correctness issue let's err on the side of less alarms.

jakkdl: Thinking about this one more, is it actually that big of a style problem to have multiple `withitem`s in a single `with`? And if the open was placed in it's own `with` it'd clearly be a TRIO112 (or TRIO302/111).

I changed it to be an error, but trivial to change it back.

jakkdl:
rebased onto main, added support for multiple withitems, cleaned up code.
Also touched on some other stuff, partially related to this. Don't think worth bothering with a separate PR for it (will only be a minor conflict with #22) but can split if you want.

flake8_trio.py
```diff
@@ -269,7 +259,7 @@ def visit_Yield(self, node: ast.Yield):
         self.generic_visit(node)
 
     # ---- 109 ----
-    def check_109(self, node: ast.AsyncFunctionDef):
+    def check_for_trio109(self, node: ast.AsyncFunctionDef):
```
jakkdl, line 262: realized function names were inconsistent, harmonized them

flake8_trio.py
```diff
@@ -333,10 +353,30 @@ def has_exception(node: Optional[ast.expr]) -> str:
 
 
 class Visitor102(Flake8TrioVisitor):
+    class TrioScope:
```
jakkdl, line 356: `get_matching_call` (previously `get_trio_scope`) no longer constructs a `TrioScope`, since only `Visitor102` actually cared about it. So moved it to a subclass

flake8_trio.py
```diff
@@ -63,32 +75,21 @@ def __eq__(self, other: Any) -> bool:
...
-                    self.shielded = kw.value.value
-                # sets to True even if timeout is explicitly set to inf
-                if kw.arg == "deadline":
-                    self.has_timeout = True
-
-    def __str__(self):
-        # Not supporting other ways of importing trio, per TRIO106
-        return f"trio.{self.funcname}"
+def get_matching_call(
+    node: ast.AST, *names: str, base: str = "trio"
```
jakkdl, line 82: A bit obscuring to default to `"trio"` when no longer having it in the function name, but probs fine.

flake8_trio.py
```diff
@@ -11,7 +11,18 @@
...
+    Dict,
+    Iterable,
+    List,
+    NamedTuple,
+    Optional,
+    Set,
+    Tuple,
+    Union,
+    cast,
+)
```
jakkdl, line 25: damn I long for when 3.8 is EOL, a pain to have to add stuff up here all the time (should look for if there's any good neovim plugins that suggests and automatically adds imports)

flake8_trio.py
```diff
@@ -290,18 +280,48 @@ def visit_Import(self, node: ast.Import):
...
+            # get variable name <X>
+            if not isinstance(item.optional_vars, ast.Name):
+                continue
+            var_name = item.optional_vars.id
+
+            # check for trio.open_nursery
+            nursery = get_matching_call(item.context_expr, "open_nursery")
+
+            # isinstance(..., ast.Call) is done in get_matching_call
+            body_call = cast(ast.Call, node.body[0].value)
```
Zac-HD, line 312: I'd use an `assert` rather than a `cast()` here.

flake8_trio.py
Zac-HD: IMO this isn't an affirmative reason *to* move it, so I'd keep the diff a bit smaller by leaving it where it was.  On the other hand, nor is there any real reason to move it *back*!

Zac-HD (approved):
LGTM!

flake8_trio.py
jakkdl: `assert` is actually not possible here, the isinstance call is done later inside `get_matching_call` (which if it inside there evaluates false propagates to the condition calling it being false).

the easiest alternative is a redundant `isinstance` in the condition

I suppose the cast is incorrect in that `body_call` isn't actually a `Call` *yet*, but is for brevity/convenience later.

flake8_trio.py
jakkdl: haha. I liked subclassing it for people reading it later on not encountering it through a commit diff, fun tension

PR #38, opened 2022-08-15
103 no longer ignore raises in else, unless there's a break

### LINKED ISSUE #35: TRIO103: Check for guaranteed raise in loops?
Reading the readme, it says
> TRIO103: except BaseException and except trio.Cancelled with a code path that doesn't re-raise. Note that any raise statements in loops are ignored since it's tricky to parse loop flow with break, continue and/or the zero-iteration case.

But having implemented that logic with 107/108 I could extend that to 103 as well, likely without too much trouble.

jakkdl:
fixes #35 

I'm not sure this really affects any code out there, but 103 should cover all theoretical cases now other than weird nested try's, afaik.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 *[CalVer, YY.month.patch](https://calver.org/)*
 
+## 22.8.6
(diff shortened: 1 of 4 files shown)


Zac-HD (approved):
Looks great!  Thanks again @jakkdl 😁

Issue #40, opened 2022-08-16
False alarm: TRIO107 for abstractmethods


Zac-HD:
```python
from abc import ABC, abstractmethod


class Handle(ABC):
    @abstractmethod
    async def ajoin(self):
        ...
```
```
$ flake8 example.py
example.py:6:5: TRIO107: exit from async function with no guaranteed checkpoint or exception since function definition on line 6
```

Note that we'll want to handle abstractclassmethod and abstractproperty too.  In addition to the explicit decorator-based handling, I'd skip 107 for functions where `...` is the complete body.

jakkdl:
how about functions with `pass` as the complete body?

jakkdl:
and if checking for `...`, is there any reason to check for the decorators explicitly?

Zac-HD:
Yeah, let's treat pass the same as ellipses. (I dislike the style, but whatever)

And we still need to handle the decorators because it's fairly common to put a `raise NotImplementedError`... actually we could handle that (and optionally with a message) the same way too. Still worth handling via decorators in case of eg `msg = "long string" ; raise NotImplementedError(msg)` or something though, which doesn't fit the one-statement heuristic.

jakkdl:
if the body is only a `raise` it's fine, since raises are treated as checkpoints.
`tests/trio107.py:248`
```python
async def foo_raise_1():  # safe
    raise ValueError()
```
and any extra statements with strings or variables or stuff won't bother that.

In theory an abstract class can contain plenty code that gets executed though, due to subclasses calling them through `super()` and the like:

```python
class Parent:
  @abstractmethod
  async def foo(self):
      for i in range(10):
          yield i

class Child(Parent):
  async def foo(self):
    return tuple(super().foo())
```
(invalid python code, but you get what I mean)

So unless you have more cases it feels like the right thing to do is ignore the decorators and just check for ellipsis or pass?

Zac-HD:
> So unless you have more cases it feels like the right thing to do is ignore the decorators and just check for ellipsis or pass?

👍

jakkdl:
Changed it to be `any statement in body that's not a string, pass or ellipsis` - and when tinkering with test cases I realized that pyright has the same check for overloaded functions `error: <> is marked as overloaded, but it includes an implementation` and when digging into it turns out they have the exact same logic ^^
[...]

PR #42, opened 2022-08-16
ignore 107/108 in asynccontextmanager, asyncWith always checkpoint on both entry and exit

### LINKED ISSUE #39: False alarm for TRIO108 in context managers
```python
from contextlib import asynccontextmanager
import trio


@asynccontextmanager
async def in_background(*background_tasks):
        for bgt in background_tasks:
            nursery.start_soon(bgt)
        yield
```
doesn't emit TRIO101 because we know this is (probably) a context manager, but we don't have the same filter for 108:
```
$ flake8 example.py 
example.py:6:1: TRIO108: exit from async iterable with no guaranteed checkpoint since yield on line 10
example.py:10:9: TRIO108: yield from async iterable with no guaranteed checkpoint since function definition on line 6
```

jakkdl:
fix #39

### DIFF
--- a/tests/trio108.py
+++ b/tests/trio108.py
@@ -544,8 +544,9 @@ async def foo_func_5():  # error: 0, "exit", Statement("yield", lineno+2)
             ...
 
 
-# Completely ignore checking asynccontextmanagers, since they checkpoint on both
-# __aenter__ and __aexit__
+# Completely ignore checking asynccontextmanagers, since they might checkpoint on
+# either or both of __aenter__ and __aexit__ and correct context managers only
+# actually yield once (even if there might be multiple yield statements).
 @asynccontextmanager
 async def foo_cm_1():
     while True:
@@ -554,6 +555,7 @@ async def foo_cm_1():
 
 @contextlib.asynccontextmanager
 async def foo_cm_2():
+    # Just imagine some fancy control flow so that only one of these executes.
     yield
     yield
     yield
@@ -565,6 +567,7 @@ async def foo_cm_3():
         yield
 
 
+# note: not asynccontextmanager, so treat this function as an async iterable.
 @anything.contextmanager
 async def foo_cm_4():  # error: 0, "exit", Stmt("yield", line+1)
     yield  # error: 4, "yield", Statement("function definition", lineno-1)
--- a/tests/trio108.py
+++ b/tests/trio108.py
@@ -1,9 +1,9 @@
 import contextlib
-import contextlib as anything
-from contextlib import asynccontextmanager, contextmanager
+from contextlib import asynccontextmanager
 from typing import Any
 
 import trio
+import trio as anything
 
 _: Any = ""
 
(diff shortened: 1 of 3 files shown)


jakkdl:
or wait, should 108 be completely ignored in contextmanagers?

Zac-HD:
> or wait, should 108 be completely ignored in contextmanagers?

Huh.  🤔 I think we probably *should* completely ignore it, since that's exactly equivalent in the (expected) case where exactly one `yield` statement is ever executed.  So *any* 108 warning we emit for a valid context manager would be a false alarm.

jakkdl:
rebased on top of #40 to avoid unnecessary merge conflicts, so you'll likely want to merge that before scrutinizing the diff.

Zac-HD (approved):
Rebased and added some comments, let's release it!

tests/trio108.py
```diff
@@ -554,6 +555,7 @@ async def foo_cm_1():
 
 @contextlib.asynccontextmanager
 async def foo_cm_2():
+    # Just imagine some fancy control flow so that only one of these executes.
     yield
     yield
     yield
```
jakkdl, line 561: Oh no, wait - I might have completely misunderstood how asynccontextmanagers work? Is `__aenter__` only called on first time the function is called, and `__aexit__` on the last time that it is? Or is `__aenter__` before every time function execution is continued?
Cause if it's the former, this PR is bad (at least the asynccontextmanager part) and should be slightly modified to only be marking this as safe
```python
@asynccontextmanager
async def foo():
  yield
```
(which without the acm would have two errors)

jakkdl: (All the fancy control flow logic for conditional execution of yields is implemented :wink: )

jakkdl: and e.g.
```python
@asynccontextmanager
async def foo():
  while True:
    yield
```
isn't an error with this PR

Zac-HD: Those methods are called exactly once each, when you enter and exit the with-block respectively (which may be well after the function call!).

I don't think it's worth linting for too-many-yields though, because there's more-precise runtime error detection.

jakkdl: Isn't that kind of the whole point of like 90% of 108 though? :P If they always just lead to false positives I'd disable them everywhere, but when the checks are written I don't see the point in explicitly disabling them for this one type of decorator. It's a trivial change to turn back on the too-many-yields check

Zac-HD: We're walking a fairly delicate balance here: it's not feasible to be precisely correct in all circumstances (for example, runtime codegen is a thing in Python!), so we're making tradeoffs between our false-alarm and missed-alarm rates.

IIRC, our current too-many-yields check is *pessimistic* about user code, and warns whenever there might be a problem - no missed alarms, but this leads to many false alarms on context-manager code.  I also just don't think that the multiple-yield thing is ~ever going to arise in practice in a context manager.

What we *could* do is just have the TRIO107 check active for async context managers; no matter what else is going on they should always checkpoint at least once before exiting (unless an exception is raised, etc.).

jakkdl: [starts writing long message]

*oh*, I once again realized the error I'm making. I was associating `yield` too hard with iterators, and not realizing how context managers actually work (I guess I've never written my own context manager) and didn't properly grasp your statements about context managers only supposing to `yield` once.

Lemme read up on how to actually write a `contextmanager` and see where I land with what if anything 107/108 should do with them. There's also a possibility of creating a separate error code for yielding twice in a `contextmanager` (async or not) since the logic is already here, though that's not conceptually a great fit for a trio-specific plugin (but could always be default off).
But now that I've invested the time into learning how to do configuration flags, it's relatively easy to tune the strictness of checks with those. And for checks such as this it could make a lot of sense to CI with reasonable flags, but once in a blue moon run a false-alarm prone checker against the code to see if it warns against anything new. Another target for this could have been whether to be pessimistic about whether `asynccontextmanagers` checkpoints both on entry and exit. 

Thanks for your patience :heart:

PR #43, opened 2022-08-16
treat abstract decorators as checkpoints, ignore functions with ellipsis as sole body

### LINKED ISSUE #40: False alarm: TRIO107 for abstractmethods

jakkdl:
Fix #40 

WIP until spec is solidified


jakkdl:
Should be good to go!

PR #44, opened 2022-08-16
handle loops that are guaranteed to loop in 103 and 107/108

### LINKED ISSUE #41: False alarm for TRIO107/108 when looping over nonempty static collections
```python
import trio

async def foo():
    for _ in [1, 2, 3]:
        await trio.sleep(0)
```
```
$ flake8 example.py
example.py:3:1: TRIO107: exit from async function with no guaranteed checkpoint or exception since function definition on line 3
```

We can do better because static lists and tuples (and maybe some other collections?) have an `.elts` attribute in the AST which allows us to check whether they're nonempty.

jakkdl:
fixes #41 

Also handles dicts, basic `range()` calls, and some fun edge cases with infinite loops and stuff.
Have not incremented version due to multiple open PR's

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 *[CalVer, YY.month.patch](https://calver.org/)*
 
+## Future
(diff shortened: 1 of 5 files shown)


Zac-HD (approved):
This is a *sweet* implementation 🤩

jakkdl:
:grin: 

Good tests :+1:

PR #50, opened 2022-08-22
allow other specified errors in test file

jakkdl:
Wrote this up while finagling with 107/108 and wanted to see how much of those issues were caused by 103. Ultimately didn't keep the added comments in the 107/108 files to minimize diff, but might as well push this for future use.
So can now do e.g.
```python
# INCLUDE TRIO103 TRIO108
try:
  ...
except: # TRIO103: 0, 'bare except'
  ...
finally:
  yield # TRIO108: [...]
```
Could then also add possibility of test file not matching an error code, in case all errors in it have the code specified.

### DIFF
--- a/tests/test_flake8_trio.py
+++ b/tests/test_flake8_trio.py
@@ -66,15 +66,15 @@ def test_eval(test: str, path: str):
             continue
 
         # get text between `error:` and (end of line or another comment)
-        k = re.findall(r"(?<=error:)[^#]*(?=#|$)", line)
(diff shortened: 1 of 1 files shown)


Zac-HD (approved):
Simple enough change, and the flexibility might be useful 👍 

(I don't have a personal view here but you've worked more with this than I have)

jakkdl:
Yeah I've thought about this for a long time and like the flexibility - would've been great to have from the start and is unlikely to do much now, but never hurts~

PR #51, opened 2022-08-22
added parametrized test of all possible permutations of try with 107

jakkdl:
When writing #49 I got paranoid about all the possible permutations and kept adding more and more tests to the file - so bashed out this one instead.
Since it creates a couple thousand tests it might be worth putting this behind another pytest mark [and not run by default?], and/or add back `xdist`.

The test added fails since it depends on the functionality of #49

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
         python -m pip install --upgrade pip setuptools tox
         python -m tox --notest --recreate -e test
     - name: Run tests
-      run: python -m tox -e test -- -n auto
+      run: python -m tox -e test
 
   release:
     runs-on: ubuntu-latest
--- a/tests/test_flake8_trio.py
+++ b/tests/test_flake8_trio.py
@@ -7,7 +7,7 @@ import sys
 import tokenize
 import unittest
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


Zac-HD:
If it runs in less than 60s with xdist, let's do that.

jakkdl:
~~Oh this runs in 13 seconds on my machine without xdist. Adding xdist actually makes it slower, up to 17 seconds. I guess every single test is so short that the overhead becomes dominating.~~
~~Comparing CI times to #49 it goes 37/41/32 -> 54/46/46
Pushing a tox with xdist I got 51/50/62~~
I tried doing the parametrization manually, and that pushed the time down to ~6s from ~13s (up from ~4s at main), with CI times at 35-40 seconds, so that clearly seems to be the way to go. It'll now only print the first fail, but that's probably an upside and if you want all errors when developing it'll be easy to modify the test.
Not the prettiest code anymore, but not worth the extra time to rejiggle stuff.

Zac-HD (approved):
Looks good!

I fixed the tox posargs handling and switched out the nested for-loop for the equally-fast `itertools.product()`; merging now 🙂

jakkdl:
ooh that's much prettier, nice!

PR #52, opened 2022-08-23
Fix trio107&108 false alarm on infinite loops

### LINKED ISSUE #47: False alarm TRIO107 with nested `while`-loops
```python
async def f():
    while True:
        if ...:
            await trio.sleep(0)
            return
        # If you delete this loop, no warning.
        while foo():
            await bar()
```
```
$ flake8 example.py
example.py:1:1: TRIO107: exit from async function with no guaranteed checkpoint
                         or exception since function definition on line 1
```

but plainly deleting the inner loop doesn't change whether the function checkpoints before exiting.

jakkdl:
So satisfying when all the tools available makes a fix close to trivial to implement.
Replaced all instances of `while ...` in the test files since `bool(...)` is True.
Fixes #47

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 # Changelog
 *[CalVer, YY.month.patch](https://calver.org/)*
 
-## Future
+## 22.8.8
 - Fix false alarm on TRIO107 with checkpointing `try` and empty `finally`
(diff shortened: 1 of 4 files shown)


Zac-HD (approved):
Looks great, thanks again!

PR #53, opened 2022-08-23
option for decorator to disable checkpoint checking

jakkdl:
flake8 for some reason freaks out about `TRIO` not being 3 characters when run through `flake8.main.application.Application()`, not sure if that's an issue somewhere on my end or in tox/pytest settings - tried updating everything in my virtualenv and wiping `.tox` but to no avail.

~~Modified the 108 check fixed in #42 to only be about entry/exit checking. Should probably be split into it's own PR, esp for ease of reverting if it does give a bunch of false alarms.~~
Use the flag to disable checking them, upcoming new PR/issue to discuss how to check. 

There's some alternatives for how to parse the disables. Should `--no-checkpoint-warning-decorators=app` disable on `app.route`, or should you require `app.*` or `app.route` (latter is current behaviour).


jakkdl:
Finished!

I'm quite surprised we haven't hit an error for the plugin code being invalid previously. The test suite never involved actually running the plugin through flake8, but surely you've been doing that? Regardless, the documentation quite explicitly says that four-letter entry points are invalid.
> Please Note: Your entry point does not need to be exactly 4 characters as of Flake8 3.0. Single letter entry point prefixes (such as the ‘X’ in the examples above) have caused issues in the past. As such, please consider using a 2 or 3 character entry point prefix, i.e., ABC is better than A but **ABCD is invalid**. A 3 letters entry point prefix followed by 3 numbers (i.e. ABC123 ) is currently the longest allowed entry point name.

https://flake8.pycqa.org/en/latest/plugin-development/registering-plugins.html

Changed the entry point to `TRI` for now, could also update all error messages to match.


Removed custom handling of `asynccontextmanager`s fully for now, you can get back previous functionality by adding it to the list of decorators disabling checks.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 
 ## Future
 - Fix false alarm on TRIO107 with checkpointing `try` and empty `finally`
+- Added `--no-checkpoint-warning-decorators`, 107/108 no longer ignores `asynccontextmanager` by default.
```
Zac-HD, line 6: I think the default value of this setting should ignore `asynccontextmanager`, so that if you don't explicitly configure anything you get the prior-to-this-PR behaviour.  This matches e.g. Coverage's ignore-patterns config; `# pragma: no cover` is ignored-by-default but if you're customizing ignore patterns you have to specify it manually.

flake8_trio.py
```diff
@@ -200,6 +205,36 @@ def has_decorator(decorator_list: List[ast.expr], *names: str):
     return False
 
 
+def regex_has_decorator(decorator_list: List[ast.expr], *names: str):
+    def wild_match(match_str: str, target: str) -> bool:
+        if match_str == "*":
+            return True
+        if "*" not in match_str:
+            return match_str == target
+        return bool(re.match(re.escape(match_str).replace("\\*", ".*"), target))
```
Zac-HD, line 214: Seems much safer to construct the fully-qualified decorator name as a string, then [use the `fnmatch` module](https://docs.python.org/3/library/fnmatch.html#fnmatch.fnmatchcase) for explicitly glob-style matching.  This more complicated logic looks like an invitation to weird edge cases.

tests/test_decorator.py
```diff
@@ -0,0 +1,121 @@
...
+    assert tuple(plugin.run())
+
+    plugin.options = Namespace(no_checkpoint_warning_decorators=["app.route"])
+    assert not tuple(plugin.run())
+
+
+def test_command_line_1(capfd):
+    Application().run(
+        [
+            "--ignore=E,F,W",
```
Zac-HD, line 85: Might prefer `--select-TRIO`?

Zac-HD (commented):
Generally looks good; more detailed comments below.

> There's some alternatives for how to parse the disables. Should `--no-checkpoint-warning-decorators=app` disable on `app.route`, or should you require `app.*` or `app.route` (latter is current behaviour).

I prefer the latter, i.e. currently-implemented behaviour, to reduce accidental suppressions👍

CHANGELOG.md
jakkdl: Ah sure, I was thinking ahead to future PR's handling `asynccontextmanager` and specifically turning off multiple-yield warnings and/or tweaking the functionality of checkpoint before `and`/`or` after yield. But future's the future

flake8_trio.py
jakkdl: hmm, that does not support differentiating `.` - so `mydeco*` would match `mydecoratormodule.foo`.
But it should be possible to get rid of the recursion and just do a `re.match` against a constructed string, which should make for simpler logic.

Zac-HD: I can see where you're coming from, but IMO the simplicity of "just use fnmatch" (both for implementation, and documentation) outweighs the slightly more precise but considerably harder to explain regex and special handling approach.

jakkdl: After playing with both I wasn't convinced at first and got the regex code simplified and readable, but then I realized that saying "ignore anything in module `x` or any of it's sub-packages" wasn't possible with simple regex logic and settled on fnmatch.

flake8_trio.py
```diff
@@ -200,6 +206,26 @@ def has_decorator(decorator_list: List[ast.expr], *names: str):
...
+    def construct_name(expr: ast.expr) -> str:
+        if isinstance(expr, ast.Name):
+            return expr.id
+        assert isinstance(expr, ast.Attribute)
+        return construct_name(expr.value) + "." + expr.attr
+
+    for decorator in decorator_list:
+        qualified_decorator_name = construct_name(decorator)
+        for pattern in patterns:
+            # remove any leading @
```
Zac-HD, line 220: `str.lstrip()` is great for this. I'd also use `any()` for the inner loop

tests/test_decorator.py
```diff
@@ -0,0 +1,120 @@
...
```
Zac-HD, line 84: --select here too?

Zac-HD (commented):
Looks good! Comments below.

flake8_trio.py
jakkdl: `lstrip` def cleaner :+1:

I don't think `any()` is cleaner here
```python
for pattern in patterns:
    if fnmatch(qualified_decorator_name, pattern.lstrip("@")):
        return True
```
vs
```python
if any(
    fnmatch(qualified_decorator_name, pattern.lstrip("@"))
    for pattern in patterns
):
    return True
```
and they should be the same efficiency

tests/test_decorator.py
jakkdl: derp, of course

Zac-HD (approved):
Awesome! I'll cut a new release next week when I'm back at a desk 😊

PR #59, opened 2022-11-22
expand 105 to also check for nursery.start

### LINKED ISSUE #56: Expand 105 to check that `nursery.start(...)` is awaited
I think we can just assume that almost all nurseries are named `nursery` and also that ~everything named `nursery` is in fact a nursery, at which point this is a pretty simple addition change:

https://github.com/Zac-HD/flake8-trio/blob/a900cffc2e47c36a9d0576cd384eb20ca33289b4/flake8_trio.py#L820-L821

```python
and (
    (node.func.value.id == "trio" and node.func.attr in trio_async_functions)
    or (node.func.value.id == "nursery" and node.func.attr == "start")
)
```

plus tests etc.

jakkdl:
Took me a second to remember what the rest of the code in the 105 checker did, so added a corresponding comment at line 812 otherwise unrelated to this PR.

fixes #56

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 *[CalVer, YY.month.patch](https://calver.org/)*
 
+## Future
+- TRIO105 now also checks for nursery.start() and nursery.start_soon()
+
 ## 22.11.1
 - TRIO102 is no longer skipped in (async) context managers, since it's not a missing-checkpoint warning.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Changelog
 *[CalVer, YY.month.patch](https://calver.org/)*
 
-## Future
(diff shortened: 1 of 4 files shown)


Zac-HD (approved):

Zac-HD:
[`nursery.start_soon()` is sync](https://trio.readthedocs.io/en/stable/reference-core.html#trio.Nursery.start_soon), so I reverted some of the more complicated stuff, and then added the new version number 🚀

jakkdl:
Ah oops, good catch!

Sure would be handy if a linter caught when you try to await a non-async function :P

Zac-HD:
Haha, at least that fails *loudly* at runtime 😄

jakkdl:
Doesn't help my tests though! Thought about running the code in the tests for a second to help check for that, but then realized there's plenty code in those that'd make the interpreter scream *very* loudly :grin:

PR #60, opened 2022-11-22
add TRIO113, nursery.start_soon() in async context manager

### LINKED ISSUE #57: When opening a context manager, prefer `await nursery.start()` over `nursery.start_soon()`
One pattern I've seen a few times lately is a context manager which ensures some external service, e.g. an external process to interact with, is available.  The code might look something like this:

```python
@asynccontextmanager
def run_sampling_profiler():
        nursery.start_soon(trio.run_process, ["start-sampling-profiler", str(os.getpid())])
        yield
```

The only problem is that the code in the calling `with`-block can start running before our external process is actually running!  To fix this, we should have used `await nursery.start(trio.run_process, ...)`.

So to *lint* for this, we should check `nursery.start_soon()` calls either inside `async def __aenter__(self):` or before the `yield` statement in a function like the above, and raise a warning if the target callable could instead be used with `nursery.start()`.  But how do we distinguish those startable targets?

- They [accept a `task_status=` keyword argument](https://trio.readthedocs.io/en/stable/reference-core.html#trio.Nursery.start)... but doing cross-module introspection would be a pretty nasty latency hit
- We can list out known cases - [search `task_status=` on this page](https://trio.readthedocs.io/en/stable/reference-io.html), and I'll be happy to expand the list if there are other suggestions
- We can support a configuration option to add more on a project-specific basis

jakkdl:
Fixes #57 

Quick and dirty:
* the message probably needs bettering.
* require that the object `start_soon` is called on is called "nursery"?
* require that it's also inside a `with`?
* could probs extend tests, but I think I covered most cases.
* maybe doesn't need it's own class, but since speed hasn't been an issue /shrug
* add config option

### DIFF
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -127,6 +127,10 @@ class Error:
     def __eq__(self, other: Any) -> bool:
         return isinstance(other, Error) and self.cmp() == other.cmp()
 
+    def __repr__(self) -> str:
+        trailer = "".join(f", {x!r}" for x in self.args)
+        return f"<{self.code} error at {self.line}:{self.col}{trailer}>"
+
 
 checkpoint_node_types = (ast.Await, ast.AsyncFor, ast.AsyncWith)
 cancel_scope_names = (
@@ -806,12 +810,13 @@ trio_async_funcs = (
 )
 
 
-def is_nursery_call(node: ast.AST, name: str) -> bool:
+def is_nursery_call(node: ast.AST, name: str) -> bool:  # pragma: no cover
     assert name in ("start", "start_soon")
     if isinstance(node, ast.Attribute):
-        if not isinstance(node.value, ast.Name):
-            return is_nursery_call(node.value, name)  # might be self.nursery.start()
-        return node.value.id.endswith("nursery") and node.attr == "start"
+        if isinstance(node.value, ast.Name):
+            return node.attr == name and node.value.id.endswith("nursery")
+        if isinstance(node.value, ast.Attribute):
+            return node.attr == name and node.value.attr.endswith("nursery")
     return False
 
 
@@ -827,8 +832,17 @@ class Visitor105(Flake8TrioVisitor):
         self.node_stack.pop()
 
     def visit_Call(self, node: ast.Call):
-        if is_nursery_call(node.func, "start") and (
-            len(self.node_stack) < 2 or not isinstance(self.node_stack[-2], ast.Await)
+        if (
+            isinstance(node.func, ast.Attribute)
+            and isinstance(node.func.value, ast.Name)
+            and (
+                (node.func.value.id == "trio" and node.func.attr in trio_async_funcs)
+                or is_nursery_call(node.func, "start")
+            )
+            and (
+                len(self.node_stack) < 2
+                or not isinstance(self.node_stack[-2], ast.Await)
+            )
         ):
             assert isinstance(node.func, ast.Attribute)
             self.error("TRIO105", node, node.func.attr)
@@ -1179,7 +1193,7 @@ class Visitor113(Flake8TrioVisitor):
     def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef):
         outer = self.aenter
 
-        self.aenter = (node.name == "__aenter__" and len(node.args.args) == 1) or any(
+        self.aenter = node.name == "__aenter__" or any(
             _get_identifier(d) == "asynccontextmanager" for d in node.decorator_list
         )
 
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -29,7 +29,7 @@ from typing import (
 from flake8.options.manager import OptionManager
 
 # CalVer: YY.month.patch, e.g. first release of July 2022 == "22.7.1"
-__version__ = "22.11.2"
+__version__ = "22.11.3"
 
 
 Error_codes = {
@@ -806,6 +806,15 @@ trio_async_funcs = (
 )
 
 
+def is_nursery_call(node: ast.AST, name: str) -> bool:
+    assert name in ("start", "start_soon")
+    if isinstance(node, ast.Attribute):
+        if not isinstance(node.value, ast.Name):
+            return is_nursery_call(node.value, name)  # might be self.nursery.start()
+        return node.value.id.endswith("nursery") and node.attr == "start"
+    return False
+
+
 class Visitor105(Flake8TrioVisitor):
     def __init__(self, *args: Any, **kwargs: Any):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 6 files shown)


jakkdl:
huh, test failures looks like it's an issue with the CI?

flake8_trio.py
```diff
@@ -65,6 +65,7 @@
         "Nurseries should generally be the inner-most context manager."
     ),
     "TRIO112": "Redundant nursery {}, consider replacing with directly awaiting the function call",
+    "TRIO113": "Dangerous `.start()`, process might not run before `__aenter__` exits. Consider replacing with `.start_soon()`.",
```
Zac-HD, line 68: This is reversed: we want to warn because `nursery.start_soon()` might not take effect during `__aenter__` and recommend `await nursery.start()`

flake8_trio.py
```diff
@@ -1146,6 +1147,52 @@ def visit_BoolOp(self, node: ast.BoolOp):
...
+    def __init__(self, *args: Any, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+        self.async_function = False
+        self.asynccontextmanager = False
+        self.aenter = False
+
+    def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef):
+        outer = self.aenter
+        if any(
+            isinstance(n, ast.Name) and n.id == "asynccontextmanager"
```
Zac-HD, line 1166: Probably also want to handle `<whatever>.asynccontextmanager` too; that'd usually but not always be `contextlib.`.

flake8_trio.py
```diff
@@ -1146,6 +1147,52 @@ def visit_BoolOp(self, node: ast.BoolOp):
...
+            for n in node.decorator_list
+        ) or (
+            node.name == "__aenter__"
+            and len(node.args.args) == 1
+            and node.args.args[0].arg == "self"
```
Zac-HD, line 1171: We don't need to require that this is `self`; it's a strong convention but not a requirement.

Zac-HD (commented):
Solid start, I think with tests and a config option to add more this will be ready 😁

flake8_trio.py
jakkdl: derp, brainfart. Fixed

flake8_trio.py
jakkdl: ah right. Do you think the `len()` check is worth keeping? Or checking that it's defined inside a class?

flake8_trio.py
```diff
@@ -1146,6 +1147,52 @@ def visit_BoolOp(self, node: ast.BoolOp):
...
+        self.generic_visit(node)
+        self.aenter = outer
+
+    def visit_Yield(self, node: ast.Yield):
+        self.aenter = False
+
+    def visit_Call(self, node: ast.Call) -> None:
+        if (
+            self.aenter
+            # and isinstance(node.func.value.id, ast.Name) and node.func.value.id == nursery ?
```
jakkdl, line 1184: Do you want this check? Could maybe reduce some false alarms, but also miss cases where nurseries aren't named `nursery`

jakkdl:
> huh, test failures looks like it's an issue with the CI?

nvm, the warning was unrelated. But I'd forgotten to push the test file :upside_down_face:

flake8_trio.py
Zac-HD: If in doubt, let's write as little code as possible - I think the len check should stay (the aenter protocol doesn't take args beyond self), but not defined in a class (I've occasionally seen functions manually attached to a class as methods).

flake8_trio.py
Zac-HD: Let's stick with the assumption that nurseries are always named nursery for now, false alarms suck and this choice is easier to find and change later.

flake8_trio.py
jakkdl: well, if we're checking `len(args.args)==1` to be paranoid about signature we should also check `args.posonlyargs`, `args.kwonlyargs`, `args.vararg`, `args.kw_defaults` and `args.defaults` are empty...

jakkdl:
Should be pretty much done now, other than what to do with the function signature on `__aenter__`

Zac-HD:
👀 looks like flake8 6.0.0 just came out with a new required argument in some logic we use for tests... fixing now.

jakkdl:
Looks good :rocket:

Zac-HD:
Found two bugs with it right away 💪 

Also looks like chasing down some similar patterns will result in follow-ups, so expect more issues next week 😅

jakkdl:
Awesome!

Haha, wonderful :grin:

flake8_trio.py
```diff
@@ -1183,6 +1251,21 @@ def add_options(option_manager: OptionManager):
...
+        option_manager.add_option(
+            "--startable-in-context-manager",
+            default="",
+            parse_from_config=True,
+            required=False,
+            comma_separated_list=True,
+            help=(
+                "Comma-separated list of method calls to enable TRIO113"
+                " warnings for."
+                " Use if you want to override the default methods the check is enabled for."
```
jakkdl, line 1263: Very fair that you didn't bother with two options - but this is no longer true, it extends the defaults.

tests/test_flake8_trio.py
```diff
@@ -353,6 +363,26 @@ def test_107_permutations():
...
+    )
+    plugin.add_options(om)
+    plugin.parse_options(om.parse_args(args=["--startable-in-context-manager=''"]))
+    default = {repr(e) for e in plugin.run() if e.code == "TRIO113"}
+
+    # and that the expected errors are given if we empty it and then extend it
+    arg = "--startable-in-context-manager=custom_startable_function"
+    plugin.parse_options(om.parse_args(args=[arg]))
+    errors = {repr(e) for e in plugin.run() if e.code == "TRIO113"} - default
+    assert errors == {repr(Error("TRIO113", 58, 8))}
```
jakkdl, line 383: and these comments no longer make sense

tests/trio113.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+# might be monkeypatched onto an instance, count this as an error too
+async def __aenter__():
+    nursery.start_soon(trio.run_process)  # error: 4
+    nursery.start_soon()  # broken code, but our analysis shouldn't crash
+    nursery.cancel_scope.cancel()
+
+
+# this only takes a single parameter ... right? :P
```
jakkdl, line 73: this comment likewise is nonsense after the argument length check on __aenter__ was removed.

jakkdl (commented):
Looking more closely there are some small issues, the doc error definitely warrants fixing.

Zac-HD:
Follow up PR please?  

The other thing it could cover is the idiom of using `[functools]partial()` to pass keyword arguments through - it'd be great to check `nursery.start_soon(partial(trio.run_process, cmd, shell=True))` by "seeing through" the call to partial.

PR #64, opened 2022-11-24
fix line endings to LF and add .gitattributes to enforce it in the future

### LINKED ISSUE #62: Line endings
in https://github.com/Zac-HD/flake8-trio/commit/490a77585f40c19b0fb06d494d4317a6e505f444 you converted all files to CRLF line endings. Was this intentional?

It's not a huge problem for me, but it makes diffs on my local machine have ugly `^M`'s and apparently the remotes are stored with LF endings so whenever I try to diff or rebase against a remote branch it's a mess. So unless you don't mind I'd like to convert all files back to LF and add a `.gitattributes` file to hopefully stop it from happening in the future. But I'll wait with it until there's no open PR's to avoid merging nightmares.
If it was intentional I'll add a .gitattributes file indicating CRLF is preferred, and see if I can get git to play nicely with that on my local machine.

jakkdl:
Fixes #62 

Glorious diff, but only thing that matters is the `.gitattributes` file.
The documentation and recommendations are quite confusing, with some proposing listing every single file extension, others saying you should set `* text=auto eol=LF`. But afaict this should be enough, and held up when testing it.
I tried explicitly creating a file with CRLF, git warns when committing it, but I can push it just fine and it stays as CRLF on my local machine - but if i curl it from github it's LF. So in case one is working on a windows machine and some tool or editor insist on files being CRLF - I think this should work out for everybody.

Will branch future PR's from this, so make sure to merge this one first.


Zac-HD (approved):
🚀 

What a glorious diff!

PR #67, opened 2022-11-24
Extend 113 to check for the method name anywhere in the first parameter

### LINKED ISSUE #63: Handle `functools.partial()` idiom for TRIO113
```python
def __aenter__(self):
    self._nursery.start_soon(trio.run_process, cmd)                       # already emits TRIO113
    self._nursery.start_soon(partial(trio.run_process, cmd, shell=True))  # we want this to emit TRIO113
```

Looking through the code, we actually already handle the `partial()` idiom everywhere else.  It'd be good to add a regression test for `TRIO112` with `partial()` though to make sure that *keeps* working.  (first proposed in https://github.com/Zac-HD/flake8-trio/pull/60#issuecomment-1326281226)

jakkdl:
Fixes #63 

So beautiful that it ended up just being a one-line diff in the actual code :sparkles:

### DIFF
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -1209,7 +1209,7 @@ class Visitor113(Flake8TrioVisitor):
             and is_nursery_call(node.func, "start_soon")
             and len(node.args) > 0
             and fnmatch_qualified_name(
-                node.args[:1],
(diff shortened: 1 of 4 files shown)


Zac-HD (approved):
Lovely!

Zac-HD:
Found and fixed a bug with this already 🎉

PR #68, opened 2022-11-25
add testing dimension for flake8=5 and flake8=6, add py311

### LINKED ISSUE #65: install_require flake8>=6.0?
Should we update `setup.py` to require flake8>=6.0, or modify tests so they can run on both <6.0 and >=6.0 - so we can make sure that code doesn't break on <6.0? Should be unlikely, but given that this one completely snuck up on me it's probably worth taking into account.

jakkdl:
fixes #65
also adds support for py311, and rejigs a bunch of the configs. Took inspiration from black's tox.ini, and got the filterwarnings from hypothesis.

crossing my fingers that the CI supports 3.11 in an easy way such as this.
Not sure this needs an entry in the changelog?
Also, looking at setup.py - did you forget to change the author_email when creating it? That sure looks like .. not your email.


Zac-HD:
Looks great - I don't think this needs a changelog entry (end-users don't need to care), and will fix the email address (oops).

The filterwarnings spec... sent me down a rabbithole, https://github.com/HypothesisWorks/hypothesis/pull/3514 and hypothesmith update tbd.    Will update, push, and merge later this evening.

PR #69, opened 2022-11-25
Add TRIO114: unknown startable

### LINKED ISSUE #66: Check that `nursery.start()`-able callables are known to flake8-trio
Our `TRIO113` check is lovely, but as soon as you go beyond Trio's built-in `.start()`-able functions it gets a lot less magical, because you have to manually configure the functions or methods which take a `task_status=` argument, and remember to update that every time you add a new one.

So... why not make the linter help with that too?  If we see an async function definition `fn` with an argument named `task_status` and a default value of `trio.TASK_STATUS_IGNORED`, and none of `fn`, `*.fn`, or `self.fn` are in the `startable_in_context_manager` list, emit a warning!

I considered adding this to `TRIO113` or making it a new warning.  Adding would mean they get atomically disabled together, but conflates semantically-kinda-distinct issues; a separate warning means we're using the same config in two places and might only check one or the other.  On net I think expanding 113 is better because it never makes sense to check only one or the other.

(On a meta-level, this is basically a hacky way to get the benefits of inter-module analysis by paying write-time rather than run-time costs; I actually kinda like it as a design pattern.  There's a continuum of designs with increasingly-automated caching here...)

jakkdl:
fixes #66 
Branched from #68 - so the diff is a bit ugly.
The manual test added should probably have a helper function.

I'm not entirely happy with the logic here, it now checks function names only against the last part of *any* specified pattern - which means that a startable function `foo` in module `bar`, that's only ever used in the code as `bar.foo`, will never prompt you to add it if there's a pattern for `other_module.foo`.
Shared function names across modules/classes is pretty common, so means you can't relax and rely on TRIO114 to remind you every time you forget a method.

Possible solutions:
1. Do a more sophisticated check, if the function name is foo, and the specified pattern is `zee.foo` check if we're in module `zee`, or we're a class method, otherwise error. If the pattern is just `foo` or `*.foo` it's satisfied.
2. Keep track of which option names, not including wildcards, have been matched previously. So if we encounter two function definitions with the name `foo`, and there's only one pattern specifying 'modulename.foo', be unhappy?

Not super happy with either of those. Could maybe make it dumber and give a flag entering a verbose mode where it will print every time a positive match is made (instead of warning when there's no match) - meant to be inspected manually by the end-user every now and then. So e.g. prints
> function `foo` in file `path/to/bar.py` matches pattern `random_other_module.foo`
That means trio114 can stay dumb and warn on obvious missing patterns, but you can every now and then run with this flag and double check that all your methods are properly matched.

(TRIO113 could similarly also do that, to catch cases where modules are imported as aliases, or class methods are passed).


Zac-HD (approved):
Looks great!

Once I get the testing-dimension PR in, I'll rebase and then merge this 😁

Zac-HD:
> I'm not entirely happy with the logic here, it now checks function names only against the last part of _any_ specified pattern - which means that a startable function `foo` in module `bar`, that's only ever used in the code as `bar.foo`, will never prompt you to add it if there's a pattern for `other_module.foo`.  Shared function names across modules/classes is pretty common, so means you can't relax and rely on TRIO114 to remind you every time you forget a method.

I'm actually OK with this; in practice all my patterns consist of either a bare `name` or `*.name`, which is handled.  More generally, the error model for TRIO114 is "contributor didn't know or forgot to add the new startable thing", which I think is still pretty well covered.  

As a backup manual check, grepping for `task_status.*=trio\.TASK_STATUS_IGNORED` and cross-referencing the config works well - I did that for my code, and then thought of this new check when considering what would happen to the list as other people add new code 🙂

Zac-HD:
✅ Found a function that my grepping had missed, though registering it didn't trigger any new TRIO113 warnings 😁

jakkdl:
> > I'm not entirely happy with the logic here, it now checks function names only against the last part of _any_ specified pattern - which means that a startable function `foo` in module `bar`, that's only ever used in the code as `bar.foo`, will never prompt you to add it if there's a pattern for `other_module.foo`.  Shared function names across modules/classes is pretty common, so means you can't relax and rely on TRIO114 to remind you every time you forget a method.
> 
> I'm actually OK with this; in practice all my patterns consist of either a bare `name` or `*.name`, which is handled. More generally, the error model for TRIO114 is "contributor didn't know or forgot to add the new startable thing", which I think is still pretty well covered.
> 
> As a backup manual check, grepping for `task_status.*=trio\.TASK_STATUS_IGNORED` and cross-referencing the config works well - I did that for my code, and then thought of this new check when considering what would happen to the list as other people add new code slightly_smiling_face

right, makes sense. I guess you're not importing startable functions across modules (without using from <x> import <y>) too much then. It might make sense to add this to the readme then as a best practice, and/or suggest using `grep`.

PR #76, opened 2022-12-06
add trio115: use trio.lowlevel.checkpoint() instead of trio.sleep(0)

jakkdl:
I think 15 minutes to write a new check is a new record!

* I opted not to require it being in an async function or be await'ed, but can change that if you want
* feel free to suggest modifications to error message or documentation as usual
* did not update `__version__` with there being multiple open PR's


Zac-HD (approved):
I'll tweak the changelog after merging locally to avoid merge conflicts; nice speedrun!

PR #78, opened 2022-12-07
import annotations from __future__ and update annotations

jakkdl:
Realized this is a thing ... why haven't we used this all along?? Managing typing imports have been a bit of a pain.

Also shed refactors all the annotations, so only manual step was adding the import at the top of the files :100: 
(and manually running shed against the `.pyi` files since new version hasn't been released)

I kinda want one of the checkers we use to suggest importing annotations from future whenever it encounters importing built-ins from typing


Zac-HD (approved):
Ah, I've habitually avoided `from __future__ import annotations` because it breaks runtime introspection for (e.g.) Hypothesis, Pydantic, and Typeguard, and at work we're on 3.10 anyway.  However runtime introspection is not really a relevant concern for flake8-trio, so let's do it!

(happily [PEP-649 will be the real future](https://discuss.python.org/t/pep-649-deferred-evaluation-of-annotations-tentatively-accepted/21331), so this will work itself out eventually)

jakkdl:
Ah, I figured there would be some kind of reason it's not used everywhere - makes sense.

ooh, fancy! Looking forward to doing `from __future__ import co_annotations` instead :smile:

Issue #58, opened 2022-11-21
Recommend async alternatives to blocking IO calls


Zac-HD:
### Background

This is going to have a fair bit of overlap with https://github.com/cooperlees/flake8-async/pull/6 and https://github.com/cooperlees/flake8-async/pull/9, precisely because I hoped for a framework-agnostic approach.  However, I've ultimately changed my mind because Trio-specific checks can be more specific, and apply to cases where there are no generic warnings or recommendations.  Using `trio.wrap_file(os.fdopen(..))` is an example of both!

### Specific recommendations

I think each cluster of these should probably use a separate `TRIO2xx` code, so they can be individually suppressed - it's useful for incremental adoption at least.  Let's reserve `TRIO200` for user-supplied recommendations (see below).

* [x] `HTTP` requests: #94 
  - `requests` and sync-`httpx` - as for `flake8-async`
  - `urllib3.request()`, but not bare `request()` (too generic) - use equivalent method with an `httpx.AsyncClient`
  - `urllib.request.urlopen()` (stdlib) - likewise, including `request.urlopen()` and `urlopen()`
  - [x] httpx.Client() #118

* [x] Process invocations: #94
  - `subprocess.Popen` and `os.popen` - `await nursery.start(trio.run_process, ...)`
  - other `subprocess` functions, `os.system` - `await trio.run_process(...)`
  - `os.spawn[vl]p?e?` family - usually as for system; if `mode=` is given and not `[os.]P_WAIT` then as for popen.

File IO:
- [x] Use `trio.open_file()` instead of `open()`, `io.open`, `io.open_code`, or `os.fdopen` in async functions #101
- [x] Use of blocking methods on file objects - wrap in `trio.wrap_file()` to get an *async* file object (see below)
- [x] Use of blocking methods on `pathlib.Path` objects - wrap in `trio.Path()` to get an async path object (again, see below) #104 #111

Path manipulation:
- [x] Any use of `os.path` module - prefer using `trio.Path` objects (it's a copout but good advice IMO) #100
- [ ] TBD: `os.walk`, `os.pidfd_open` - don't have exact equivalents on `trio.Path`? 🤔 

OS functionality:
- `os.wait.*()` (except `waitstatus_to_exitcode`) - must be wrapped in `trio.to_thread.run_sync()`
- `os.pidfd_open` - must be wrapped in `trio.wrap_file()` immediately
[...]

PR #80, opened 2022-12-08
Add TRIO200, user-configured warnings for blocking sync calls in async function

jakkdl:
Implements the config option in #58 
This ended up touching surprisingly many spots in various places, most of them unintentionally. So sorry for the diff being a bit sprawling. Going through the diff and explaining all changes:

[flake8_trio.py]
* Currently it works with wildcards, only needed a minor modification to `fnmatch_qualified_name` for that to work smoothly, which doesn't affect the decorator matching using it.
* Visitor200 is pretty simple
* Couldn't find any built-in argparse dict parser, so wrote one myself. I would maybe have used `:` as the splitter if it was me, but easily configurable in `ParseDict`.
  * Considered splitting into two separate argumenterrors for no splitter tokens and >1 tokens, but idk if users need that much hand-holding
* after you taught me about `f{var!r}` I couldn't resist adding it to `ListOfIdentifiers` as well ^^;

[pyproject.toml]
* got tired of managing pyright complaining about typing errors in my test file, so added an exclude filter (and needed to explicitly add the default excludes). This means that a bunch of other test files could be cleaned up
  * Was there any reason this wasn't done from the start?

[tests/test_decorator.py]
* test_decorator started breaking with another command-line option, so modified that to use the same construction as the `test_xxx_options` in `test_flake8_trio.py`

[tests/test_flake8_trio.py]
* with the increasing number of warnings with configuration flags, and realizing we're back to manually specifying error codes and lines and shit, I added support for an `# ARGS` construct in a test file to specify command line flags to be used when running the file.
  * No obvious way to check malformed flags with that approach, so that's still in a separate test function
  * But `test_113_options` and `test_114_options` could be cleaned up given this new functionality. Leaving that for a later PR though.
* I wanted capsys functionality in `test_200_options`, partly because testing is good:tm:, but in large part because I wanted to silence them in my CLI even when running with `-s`. So added that to `test_114_options` as well.

TODO:
* CHANGELOG entry :heavy_check_mark: 
* README entry :heavy_check_mark: 
* README configuration flag documentation :heavy_check_mark: 
* The test file could maybe have some comments :heavy_check_mark:


jakkdl:
sad, I wrote and confirmed that a version with `ast.unparse`, removing `construct_name` completely, works perfectly with current tests - until 3.8 came and spoiled all the fun.

tests/trio200.py
```diff
@@ -0,0 +1,22 @@
+# ARGS --trio200-blocking-calls=bar->BAR,bee->NOERROR,bonnet->NOERROR,bee.bonnet->BEEBONNET
```
Zac-HD, line 1: is NOERROR special?

Zac-HD:
Looks good so far!  How much would break if we just declared that 3.8 was only best-effort support?  I don't really mind requiring 3.9+ to run the linter, now that 3.11 is out.

My (weakish) preference for `->` over `:` is that the latter often has special meaning in config file formats, or can be confused with attribute/member lookup from other languages.  `->` to me reads more directly as "replace with" but I'm open to arguments against!

jakkdl:
For just this change it would break 107,108 & 200 - and almost certainly all other 2xx messages. But there's generally some developing overhead to account for 3.8, since the AST has changed since then (e.g. https://github.com/PyCQA/flake8-bugbear/blob/main/bugbear.py#L683-L689) so it can sometimes be a non-insignificant overhead to debug issues caused by it, and even if that's just a matter of skipping broken tests you'd still have to figure out where/why/how it breaks and insert skips, and also update tests to skip it. So I personally wouldn't complain if we stopped testing on it entirely (and running the full test suite would be faster, yay!).

jakkdl:
> My (weakish) preference for `->` over `:` is that the latter often has special meaning in config file formats, or can be confused with attribute/member lookup from other languages. `->` to me reads more directly as "replace with" but I'm open to arguments against!
Makes sense! I'm mostly just colored by mentally parsing it as a dict, and seeing `->` just looks a bit weird. But since we don't require `{}` so it's not a proper dict it's maybe a bad idea to reinforce the idea that it's parsed into one under the hood.

tests/trio200.py
jakkdl: or actually no that could in theory cause some other confusion/testing error since BEE and BONNET are substrings of BEEBONNET.
Maybe all magic words should be surrounded by underscores or something, not sure.

jakkdl:
fixed todo's, changed `fnmatch_qualified_name` and started messing with disabling tests for 3.8 - but now it fails because of test coverage, so would need to configure `--no-cov` only when checking py38...

flake8_trio.py
```diff
@@ -202,28 +204,21 @@ def has_decorator(decorator_list: list[ast.expr], *names: str):
...
-        elif isinstance(expr, ast.Attribute):
-            attr = construct_name(expr.value)
-            assert attr is not None
-            return attr + "." + expr.attr
-        # See https://peps.python.org/pep-0614/ - we don't handle everything
-        return None  # pragma: no cover  # impossible on Python 3.8
-
+def fnmatch_qualified_name(name_list: list[ast.expr], *patterns: str) -> str | None:
+    if sys.version_info[:2] < (3, 9):
+        return None  # pragma: no cover
```
Zac-HD, line 209: Given that this is causing us coverage issues, affects 107/108 too, and I'd expect it to be trouble w.r.t. unused-ignores config across versions, let's just drop support for Python 3.8 entirely.  Drop it from the CI matrix, update the requires_python metadata, and pass `shed --py39-plus` in our tox.ini.

flake8_trio.py
```diff
@@ -1290,22 +1285,75 @@ def visit_Call(self, node: ast.Call):
...
+class ParseDict(argparse.Action):
+    def __call__(
+        self,
+        parser: argparse.ArgumentParser,
+        namespace: argparse.Namespace,
+        values: Sequence[str] | None,
+        option_string: str | None = None,
+    ):
+        res: dict[str, str] = {}
+        splitter = "->"
```
Zac-HD, line 1341:
```suggestion
        splitter = "->"  # Avoid ":" because it's part of '.ini file syntax
```

flake8_trio.py
```diff
@@ -1355,6 +1403,17 @@ def add_options(option_manager: OptionManager):
...
         )
+        option_manager.add_option(
+            "--trio200-blocking-calls",
+            default={},
+            parse_from_config=True,
+            required=False,
+            comma_separated_list=True,
+            action=ParseDict,
+            help=(
+                "Comma-separated list of key:value pairs, where key is a [dotted] function that if found inside an async function will raise TRIO200, suggesting it be replaced with {value}"
```
Zac-HD, line 1414: split long line

flake8_trio.py
```diff
@@ -1290,22 +1285,75 @@ def visit_Call(self, node: ast.Call):
...
+        values: Sequence[str] | None,
         option_string: str | None = None,
     ):
         assert values is not None
         assert option_string is not None
         for value in values:
             if not value.isidentifier():
-                raise argparse.ArgumentError(self, f"{value} is not a valid identifier")
+                raise argparse.ArgumentError(
+                    self, f"{value!r} is not a valid identifier"
```
Zac-HD, line 1327: `if keyword.iskeyword(value) or not value.isidentifier():` and `f"{value!r} is not a valid variable name"` - we don't want people putting in `True`, for example.

Zac-HD (commented):
Looking good!

README.md
```diff
@@ -66,8 +66,36 @@ Comma-separated list of methods which should be used with `.start()` when openin
...
+Comma-separated list of pairs of values separated by `->` (optional whitespace stripped), where the first is a pattern for a call that should raise an error if found inside an async function, and the second is what should be suggested to use instead. It uses fnmatch as per [`no-checkpoint-warning-decorators`](#no-checkpoint-warning-decorators) for matching. The part after `->` is not used by the checker other than when printing the error, so you could add extra info there if you want.
+
+The format of the error message is `User-configured blocking sync call {0} in async function, consider replacing with {1}.`, where `{0}` is the pattern the call matches and `{1}` is the suggested replacement.
+
+Example:
+```
+trio200-blocking-calls =
+  my_blocking_call -> async.alternative
+  module.block_call -> other_function_to_use
+  common_error_call -> alternative(), but sometimes you should use other_function(), ask joe if you're unsure which one
```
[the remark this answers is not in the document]
jakkdl: oh derp, ofc.

Zac-HD (approved):
Looks great despite growing somewhat in the writing - thanks @jakkdl!

PR #82, opened 2022-12-16
move error codes into visitor classes

jakkdl:
also changes test_113 and 114_options to use ARGS (since they broke from this it was a good opportunity to rewrite them).

> As a follow-up, it'd be nice for each visitor class to have a docstring briefly explaining what they do - just copying the relevant line from the readme would be plenty.

Yeah this has been sorely needed for a while, I started copy-pasting but started feeling very silly about it so the visitor classes now define their error codes themselves.
Would be fun to move the documentation from the readme into them and auto-generate docs (+ the same for command line args) as well, but that's probs overkill - although having to keep documentation correct in several different files & places is always a pain.

### DIFF
--- a/tests/test_flake8_trio.py
+++ b/tests/test_flake8_trio.py
@@ -21,7 +21,7 @@ from flake8.options.manager import OptionManager
 from hypothesis import HealthCheck, given, settings
 from hypothesmith import from_grammar, from_node
 
-from flake8_trio import Error, Error_codes, Plugin, Statement
+from flake8_trio import Error, Flake8TrioVisitor, Plugin, Statement
 
 trio_test_files_regex = re.compile(r"trio\d\d\d(_py.*)?.py")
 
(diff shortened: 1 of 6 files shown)


tests/test_flake8_trio.py
```diff
@@ -123,14 +130,15 @@ def test_eval(test: str, path: str):
             ), f'invalid column "{col}" @L{lineno}, in "{line}"'
 
             # assert col.isdigit(), f'invalid column "{col}" @L{lineno}, in "{line}"'
```
Zac-HD, line 132: Might be good to either enable or delete this check?

Zac-HD (approved):
🚀

tests/test_flake8_trio.py
jakkdl: Agh, high time to add flake8-eradicate to my local pre-commit.

Zac-HD: [Here's Hypothesis' list of flake8 plugins I like](https://github.com/HypothesisWorks/hypothesis/blob/226028dcf62b7c7a239574af6f5d69559be0febc/requirements/tools.in#L5-L19), fyi.

jakkdl: ooh, will go through those!

Issue #84, opened 2022-12-17
`trio200-blocking-calls =` is loaded but not parsed from a config file


Zac-HD:
I get an error on line 1353 because it hasn't been parsed from list-of-strings into a dict:

https://github.com/Zac-HD/flake8-trio/blob/7cdd9180e3aefca89d1a58d197022f266b45868c/flake8_trio.py#L1350-L1356

I also started on a quick regression test, but don't have time to finish it right now: https://github.com/Zac-HD/flake8-trio/compare/config-error

jakkdl:
I think the actual problem here is flake8's `comma_separated_list` .. apparently treating spaces as commas?
```bash
flake8 --no-checkpoint-warning-decorators="a b, c d" <file>
``` 
```python
ipdb> self.options.no_checkpoint_warning_decorators
['a', 'b', 'c', 'd']
```

jakkdl:
New try: It looks like the `action` parameter to `flake8.options.manager.OptionManager.add_option()` is entirely ignored when parsing a config file. The supplied function is never called, and it's the same thing with `ListOfIdentifiers` for `--startable-in-context-manager`, invalid identifiers aren't flagged. Instead the values are just accepted straight up.
Passing the same values at the command line works as expected.

Will delve more into flake8's internals to figure out what's going on, and probs open an issue over there. I'm opening a PR with a dirty workaround, travelling to family for holidays in the coming days so a more thorough exploration might be delayed.

(and then also fix spaces)

jakkdl:
Opened https://github.com/PyCQA/flake8/issues/1770 and https://github.com/PyCQA/flake8/issues/1771

Zac-HD:
Apparently we should just avoid `action` entirely and use `type` instead https://github.com/PyCQA/flake8/issues/1770#issuecomment-1358067558 🤷

PR #86, opened 2022-12-18
Config error

jakkdl:
Very dirty workaround for #84 in case you want it fixed quickly.
Spaces in the "suggested replacement", or around `->`, are still broken.

Also noticed a typo in the readme


Zac-HD (approved):
Heh, we'll definitely want a more principled approach eventually but this patch looks like a solid improvement on the status quo 👍 

Thanks for the quick turnaround!

PR #87, opened 2022-12-19
eradicate commented out code

jakkdl:
Also some false alarms that I ended up wrapping in "\`\`"


jakkdl:
I haven't added it to `tox.ini` due to it not being compatible with flake8-6 https://github.com/wemake-services/flake8-eradicate/issues/269

Zac-HD (approved):

PR #88, opened 2022-12-19
add pre-commit config

jakkdl:
It bit me a bunch of times that I didn't run tox before pushing minor changes, and `tox -e check` takes 13s on my system, so I've gradually recreated all the functionality from it in pre-commit + some other checks. So at this point it might be worth removing `tox -e check` entirely and transitioning the CI to use pre-commit as well. If you approve of the idea I'll implement the necessary gh CI changes as well & clean up `tox.ini`.

~~(though it currently won't pass until #87 is merged)~~

TODO:
- [x] Go through https://github.com/HypothesisWorks/hypothesis/blob/226028dcf62b7c7a239574af6f5d69559be0febc/requirements/tools.in#L5-L19 and crib any nice ones.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,67 @@
+ci:
+  skip: [pyright]
+
+repos:
+-   repo: https://github.com/Zac-HD/shed
+    rev: 0.10.8
+    hooks:
+    -   id: shed
+        args: ['--py39-plus']
+
+-   repo: https://github.com/RobertCraigie/pyright-python
+    rev: v1.1.286
+    hooks:
+    -   id: pyright
+        entry: env PYRIGHT_PYTHON_FORCE_VERSION=latest pyright
+        args: ['--pythonversion=3.11', '--warnings']
+        additional_dependencies:
+          # Required for pyright strict mode
+          - hypothesis
+          - hypothesmith
+          - pytest
+          - flake8
+
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.4.0
+    hooks:
+    -   id: trailing-whitespace
+        args: ['--markdown-linebreak-ext=md,markdown']
+    -   id: end-of-file-fixer
+    -   id: fix-encoding-pragma
+        args: [--remove]
+    -   id: check-yaml
+    -   id: debug-statements
+        language_version: python3
+
+-   repo: https://github.com/PyCQA/flake8
+    rev: 6.0.0
+    hooks:
+    -   id: flake8
+        args: ["--exclude", ".*,tests/trio*.py"]
+        language_version: python3
+        additional_dependencies:
+          - flake8-builtins
+          - flake8-bugbear
+          - flake8-comprehensions
+          - flake8-2020
+          - flake8-bandit
+          - flake8-builtins
+          - flake8-bugbear
+          - flake8-comprehensions
+          - flake8-datetimez
+         #- flake8-docstrings
+          - flake8-mutable
+          - flake8-noqa
+          - flake8-pie
+          - flake8-pytest-style
+          - flake8-return
+          - flake8-simplify
+
+-   repo: https://github.com/PyCQA/flake8
+    rev: 5.0.4
+    hooks:
+    -   id: flake8
+        args: ["--exclude", ".*,tests/trio*.py", "--select=E800"]
+        language_version: python3
+        additional_dependencies:
+          - flake8-eradicate
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -274,12 +274,11 @@ class Flake8TrioVisitor(ast.NodeVisitor):
 # ignores module and only checks the unqualified name of the decorator
 # used in 101 and 107/108
 def has_decorator(decorator_list: list[ast.expr], *names: str):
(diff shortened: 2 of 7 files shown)


Zac-HD:
Sounds good to me; mark ready-to-review when... you know... and I'll be happy to merge.

jakkdl:
I added https://pre-commit.ci/ to my fork, and it now passes on my mirror PR over there: https://github.com/jakkdl/flake8-trio/pull/1

The largest issue was that pyright [requires](https://github.com/RobertCraigie/pyright-python/issues/2) internet connectivity to run, which the pre-commit ci [does not support]( https://github.com/pre-commit-ci/issues/issues/47#issuecomment-804481763)

There's a couple options to get around that:
* Don't run pyright in pre-commit, but have a tox action for it. I'd prefer not to do this, since I want local pre-commit hooks to still run pyright when committing.
* [current solution] skip pyright when run in ci, but have a separate tox environment that runs `pre-commit run pyright --all-files`. This environment is not included in the default envlist.
  * One downside is that pre-commit will still initialize all hook environments, even if only a single hook is specified, so there's a slight time penalty here.
  * This will necessarily have to check all files (unless the modified file list is piped from the gh ci action into tox somehow) so will be slower - but this is probably an upside since an unmodified file can depend on a type specified in a modified file.
* Use https://pre-commit.ci/lite.html instead, so pre-commit is run as a GH action. Then it will have internet connectivity, but losing out on speed and automatic PR's that update hooks.
* Use mypy instead of pyright :wink: (idk why you went with pyright from the beginning actually, but since I've had mypy enabled in my editor since previously it runs perfectly fine in mypy as well - though not with `--strict` (but most of those are function signatures not having `-> None` which pyright doesn't require so it's been inconsistently applied)


Various other considerations for this, or future PR's, or never:
* Run minimal tox from pre-commit, so developers (me) don't accidentally push code with broken tests. Don't think it happens enough to bother figuring out the piping.
* Don't run tests with coverage on any platform but py311-flake8_6. Running with coverage slows down considerably, and means extra `# noqa`'s (I've hacked together some ugly aliases that mostly solves this locally)
* flake8-eradicate still doesn't work with flake8-6, and I haven't found a way to tell pre-commit to exclude specific hook id's from autoupdating, so with the full CI service we'll likely have it nagging us that flake8 is out of date.
* Do we want to disable `autofix_prs`? I haven't used pre-commit ci enough to know how handy vs irritating vs dangerous it is, but noticed that flake8-bugbear has it disabled.


But otherwise I think this should be good for merging

jakkdl:
rebased on top of main

flake8_trio.py
```diff
@@ -612,17 +611,18 @@ def visit_For(self, node: ast.For | ast.While):
         if not self.unraised:
             return
 
+        # the following block is duplicated in Visitor107_108
         infinite_loop = False
         if isinstance(node, ast.While):
             try:
                 infinite_loop = body_guaranteed_once = bool(ast.literal_eval(node.test))
-            except Exception:
+            except ValueError:
```
Zac-HD, line 619: IIRC it's *possible* to get `TypeError` on some versions?  Could be wrong, but I'd leave it as `Exception` here and below because it's a small block and I don't really care what kind of exception this raises.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,67 @@
...
+    hooks:
+    -   id: shed
+        args: ['--py39-plus']
+
+-   repo: https://github.com/RobertCraigie/pyright-python
+    rev: v1.1.286
+    hooks:
+    -   id: pyright
+        entry: env PYRIGHT_PYTHON_FORCE_VERSION=latest pyright
+        args: ['--pythonversion=3.11', '--warnings']
```
Zac-HD, line 16: I forget the exact syntax here, but we'll want this to run on all-files rather than passing changed filenames, because changing one file can create typecheck errors in another unchanged file.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,67 @@
...
+          - flake8-bugbear
+          - flake8-comprehensions
+          - flake8-datetimez
+         #- flake8-docstrings
+          - flake8-mutable
+          - flake8-noqa
+          - flake8-pie
+          - flake8-pytest-style
+          - flake8-return
+          - flake8-simplify
```
Zac-HD, line 58: 🚀 

(makes my pin-your-deps senses tingle, but no worse than the status quo I suppose... I really want better transitive-pins support from pre-commit but it's a pretty big project)

Zac-HD (commented):
Two small actionable comments, then ready to merge!

.pre-commit-config.yaml
jakkdl: In the CI check it will run on all files (it runs `pre-commit run pyright --all-files`), but I personally prefer the commit hook not to run on all files as that comes with a decently hefty time penalty (5->10 seconds on my machine).

flake8_trio.py
jakkdl: sounds reasonable, fixed :+1:

jakkdl:
Renamed the `check` ci action to show that it's now *solely* for type-checking. I tried doing that previously but failed, but idk what I did different that time.

Zac-HD (approved):

PR #90, opened 2022-12-19
don't raise TRIO200 if the call is immediately awaited.

### LINKED ISSUE #85: Don't raise `TRIO200` if the call is immediately awaited
One thing I just tried was configuring `websocket.*` as a known-sync call, in order to suggest https://pypi.org/project/trio-websocket/ over https://pypi.org/project/websocket/

But then it turns out that quite naturally we've used `websocket` as a variable name for a... websocket... and then we get a lint error on e.g. `message = await websocket.get_message()`.  One option is to go through and rename each such instance to e.g. `ws`, but another would be to *avoid* emitting `TRIO200` if the call is immediately awaited, because in these cases it's clearly *not* a blocking sync call after all.

jakkdl:
Fixes #85

Maybe nicer to rename the variable as well, but didn't immediately come up with a great name and class is small enough it won't really matter.

### DIFF
--- a/tests/trio200.py
+++ b/tests/trio200.py
@@ -4,6 +4,7 @@
 # don't error in sync function
 def foo():
     bar()
+    bee.bonnet()
 
 
 async def afoo():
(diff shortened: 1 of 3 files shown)


tests/trio200.py
```diff
@@ -39,3 +39,11 @@ async def bar():
             bar()  # TRIO200: 12, "bar", "BAR"
 
     bar()  # TRIO200: 4, "bar", "BAR"
+
+    # don't error on awaited expressions
+    # https://github.com/Zac-HD/flake8-trio/issues/85
+    await bar()
+    await print(bar())
```
Zac-HD, line 46: Ah, this actually *should* be an error, since you need to `print(await bar())`.  The tricky part is that we only want to skip the check if the _parent node_ of the call is an `await`, not _any ancestor_.  Consider e.g.

```python
await async_fn(
    value = blocking_fn()  # Want trio200: use `value = await nonblocking_fn()` here!
)
```

jakkdl: Ah, right. Fixed :+1:

jakkdl:
Made a minor runner change that also cleaned up 105, which also checks whether calls are directly awaited.

There's a couple false alarms, but I'm not sure if any of them are more than just theoretical false alarms. But if you encounter any of those, or others, it should be fairly easy to fix.

PR #92, opened 2022-12-28
fix flake8-noqa compatibility

jakkdl:
*Huge* PR :grin: 

Not sure why that colon was added in the first place, but the plugin now finally works with flake8-noqa's `NQA102` and `NQA103`


jakkdl:
I could add a regression test for this if you want.

Zac-HD (approved):
Looks good, thanks for the catch!  😅

Issue #89, opened 2022-12-19
Handle spaces in config options


jakkdl:
There's a bunch of different ways of handling spaces in comma_separated_list to fix `--trio200-blocking-calls`. Pending on what happens with https://github.com/PyCQA/flake8/issues/1771

* handle `['a', '->', 'b']` as well as `['a->b']` with fancy logic
* don't use `comma_separated_list` and write a custom action (more attractive if https://github.com/PyCQA/flake8/issues/1770 is fixed)
* Require users to wrap values in quotes or something. Though some quick testing looks like this doesn't help.
* Wait for it to be fixed upstream

jakkdl:
Since 1771 was closed as invalid, this should likely be ignored for `--startable-in-context-manager` (since spaces aren't valid identifiers anyway), but `--trio200-blocking-calls` will need a custom `type` that handles this.

Zac-HD:
Closed by #93.  (you need a keyword *per issue #*, annoyingly)

PR #93, opened 2022-12-29
config options handle spaces and newlines

### LINKED ISSUE #84: `trio200-blocking-calls =` is loaded but not parsed from a config file

jakkdl:
fix #84 and #89
`--startable-in-context-manager` and `--trio200-blocking-calls` now handle spaces and newlines.


tests/test_flake8_trio.py
```diff
@@ -90,9 +93,7 @@ def test_eval(test: str, path: str):
 
         # add command-line args if specified with #ARGS
         elif reg_match := re.search(r"(?<=ARGS).*", line):
-            for arg in reg_match.group().split(" "):
-                if arg.strip():
-                    parsed_args.append(arg.strip())
+            parsed_args.append(reg_match.group().strip())
```
jakkdl, line 96: Changed this to handle spaces in a command line arg - this means that the name `ARGS` is misleading but that is renamed in #91 (rip me bundling changes again).
This leads to some minor merge conflicts, but I'll likely rebase #91 on top of this one to fix those.

tests/test_flake8_trio.py
```diff
@@ -465,12 +467,42 @@ async def foo():
...
+    main(
+        argv=[
+            str(tmp_path / "example.py"),
+            "--append-config",
+            str(tmp_path / ".flake8"),
+        ]
+    )
+    out, err = capsys.readouterr()
+    assert not err
+    assert err_msg == out
```
jakkdl, line 498: almost funny after the fact how massively the `subprocess` call broke my debugging workflow and made finding out why tests broke take ages (a trailing comma in the list of config options).

I broke this out as a separate test in case flake8 internals break, but feels like there should be a cleaner solution that doesn't rely on any internal structure at all and just emulates calling it with `python -m`. Something like `runpy.run_module` but with also setting `sys.argv` and working dir. I didn't find anything that didn't require writing a bunch of extra code, but if you have anything I'd love to hear of it.

Zac-HD: No specific advice sorry, but I appreciate having this test broken out!

tests/test_flake8_trio.py
```diff
@@ -465,12 +467,42 @@ async def foo():
...
+
+def test_200_from_config_subprocess(tmp_path: Path):
+    err_msg = _test_trio200_from_config_common(tmp_path)
     res = subprocess.run(["flake8"], cwd=tmp_path, capture_output=True)
     assert not res.stderr
-    assert res.stdout == (
-        b"./example.py:5:5: TRIO200: User-configured blocking sync call sync_fns.* "
-        b"in async function, consider replacing with the_async_equivalent.\n"
-    )
+    assert res.stdout == bytes(err_msg, "ascii")
```
Zac-HD, line 505:
```suggestion
    assert res.stdout == err_msg.encode("ascii")
```

Semantically identical in this case, but I prefer `str.encode()` for clarity as bytes() is also callable on e.g. a list of ints.

tests/test_flake8_trio.py
Zac-HD: Sounds good - I'll next have internet around Jan 9th, so don't panic if review is a bit delayed.

Zac-HD (approved):

tests/test_flake8_trio.py
jakkdl: gtk, I'll rebase and fix merge conflicts in it. May also split it up if I start messing with stuff in other PR's that might lead to conflicts.

PR #94, opened 2022-12-31
21X and 22X, blocking sync calls in async function

jakkdl:
Checking off the first couple sync calls in #58 

`Flake8TrioVisitor.__subclasses__` didn't do it anymore once I started using multilevel inheritance, so solved it by creating a decorator for error classes that registers them. (type hinting the decorator function was fun :sweat_smile:)

Wasn't sure if the http calls and process invocation bullet points wanted different error messages for each, and the messages themselves are kinda rough, but went with this for a first pass.

TODO:
- [x] changelog
- [ ] better readme text & error messages
- [x] os.spawn[vl]p?e? should give TRIO220 sometimes
- [x] test file shouldn't be called `trio210.py` - change regex so it can be named `trio2xx.py` or split into `trio21x.py` and `trio22x.py`

Don't bother merging both this and #91, merge one and I'll fix the conflicts in the other.

### DIFF
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -18,14 +18,14 @@ import tokenize
 from argparse import ArgumentTypeError, Namespace
 from collections.abc import Iterable
 from fnmatch import fnmatch
-from typing import TYPE_CHECKING, Any, NamedTuple, Union, cast
+from typing import TYPE_CHECKING, Any, NamedTuple, TypeVar, Union, cast
 
 # guard against internal flake8 changes
 if TYPE_CHECKING:
     from flake8.options.manager import OptionManager
 
 # CalVer: YY.month.patch, e.g. first release of July 2022 == "22.7.1"
-__version__ = "22.12.5"
+__version__ = "23.1.1"
 
 
 class Statement(NamedTuple):
@@ -103,7 +103,7 @@ class Flake8TrioRunner(ast.NodeVisitor):
 
         self.visitors = {
             v(options, self._problems)
-            for v in Flake8TrioVisitor.__subclasses__()
+            for v in ERROR_CLASSES
             if self.selected(v.error_codes)
         }
 
@@ -271,6 +271,16 @@ class Flake8TrioVisitor(ast.NodeVisitor):
             yield from ast.walk(b)
 
 
+ERROR_CLASSES: set[type[Flake8TrioVisitor]] = set()
+
+T = TypeVar("T", bound=Flake8TrioVisitor)
+
+
+def error_class(error_class: type[T]) -> type[T]:
+    ERROR_CLASSES.add(error_class)
+    return error_class
+
+
 # ignores module and only checks the unqualified name of the decorator
 # used in 101 and 107/108
 def has_decorator(decorator_list: list[ast.expr], *names: str):
@@ -309,6 +319,7 @@ cancel_scope_names = (
 )
 
 
+@error_class
 class Visitor100(Flake8TrioVisitor):
     error_codes = {
         "TRIO100": (
@@ -328,6 +339,7 @@ class Visitor100(Flake8TrioVisitor):
     visit_AsyncWith = visit_With
 
 
+@error_class
 class Visitor101(Flake8TrioVisitor):
     error_codes = {
         "TRIO101": (
@@ -403,6 +415,7 @@ def critical_except(node: ast.ExceptHandler) -> Statement | None:
     return None
 
 
+@error_class
 class Visitor102(Flake8TrioVisitor):
     error_codes = {
         "TRIO102": (
@@ -505,6 +518,7 @@ class Visitor102(Flake8TrioVisitor):
 
 # Never have an except Cancelled or except BaseException block with a code path that
 # doesn't re-raise the error
+@error_class
 class Visitor103_104(Flake8TrioVisitor):
     error_codes = {
         "TRIO103": "{} block with a code path that doesn't re-raise the error",
@@ -722,6 +736,7 @@ def is_nursery_call(node: ast.AST, name: str) -> bool:
     return False
 
 
+@error_class
 class Visitor105(Flake8TrioVisitor):
     error_codes = {
         "TRIO105": "trio async function {} must be immediately awaited",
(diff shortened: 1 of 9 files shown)


jakkdl:
Not gonna bother composing beautiful error messages and readme texts until the split has been decided on.

flake8_trio.py
```diff
@@ -1421,9 +1448,89 @@ def visit_AsyncFunctionDef(
...
+        if key := fnmatch_qualified_name([node.func], *blocking_calls):
+            self.error(node, key, blocking_calls[key])
+
+
+@error_class
+class Visitor21X(Visitor200):
+    error_codes = {
+        "TRIO210": "Sync HTTP call {} in async function should use httpx.AsyncClient",
+        "TRIO211": "Sync HTTP call {} in async function should use httpx.AsyncClient",
+        "TRIO212": "Sync HTTP call {} in async function should use httpx.AsyncClient",
```
Zac-HD, line 1464: I think if we're giving the same error message, that's a signal that it should be a single error code.

Imprecise heuristics are something of an exception since they should be separately disablable... but not really because the message should also make it clear we're guessing.  (see below)

flake8_trio.py
```diff
@@ -1421,9 +1448,89 @@ def visit_AsyncFunctionDef(
...
+            return (isinstance(arg, ast.Attribute) and arg.attr == "P_WAIT") or (
+                isinstance(arg, ast.Name) and arg.id == "P_WAIT"
+            )
+
+        func_name = ast.unparse(node.func)
+        if func_name in ("subprocess.Popen", "os.popen"):
+            self.error(node, func_name, error_code="TRIO220")
+            return
+
+        if func_name in "os.system" or func_name.startswith("subprocess."):
```
Zac-HD, line 1513: Not if we construct a `subprocess.CompletedProcess` object, or a `subprocess.CalledProcessError`, etc.  Needs to be an explicit list, sadly.

flake8_trio.py
jakkdl: yeah it was more 22X that got me going on separate error codes, so I figured I'd make them separate for the moment. But esp when it's only the originating library that differs, and users aren't really gonna use too many different ones and care about the distinction within async code, these can probably all be collapsed.

Should the code be 210 or 201 (moving 22x to 21x)?

flake8_trio.py
```diff
@@ -1421,9 +1448,89 @@ def visit_AsyncFunctionDef(
...
+            "request",
+            "send",
+        )
+        for http_package in "requests", "httpx":
+            if get_matching_call(node, *http_methods, base=http_package):
+                self.error(node, ast.unparse(node.func), error_code="TRIO210")
+                return
+
+        func_name = ast.unparse(node.func)
+        if func_name == "urllib3.requests":
```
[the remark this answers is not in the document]
jakkdl: so simply match on
```python
isinstance(arg, ast.Attribute) and arg.attr == 'request'
```
and error? Or how do you want to care for "GET"/"POST"

flake8_trio.py
Zac-HD: Let's go with 21x: I think it's useful to have the middle digit group like warnings (e.g. "http requests"), and the last to distinguish specific messages/responses/confidence levels/etc.

flake8_trio.py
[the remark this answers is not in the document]
Zac-HD: That plus "the first argument is a string literal which is in GET/POST/PUT/etc".  This could still be a false alarm, or miss e.g. `urllib3.request(method_variable, ...)`, so we should keep the existing checks as-is and put this under a new error code.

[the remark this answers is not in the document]
jakkdl: What about
```python
client = httpx.Client()
client.get('https://example.org')
```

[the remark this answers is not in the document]
jakkdl: > That plus "the first argument is a string literal which is in GET/POST/PUT/etc". This could still be a false alarm, or miss e.g. `urllib3.request(method_variable, ...)`, so we should keep the existing checks as-is and put this under a new error code.

oh and we can detect urllib3 import to make this less noisy, same with httpx.

jakkdl:
addressed comments, rebased on top of main. This PR is turning out quite meaty indeed

flake8_trio.py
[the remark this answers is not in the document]
Zac-HD: > What about
> 
> ```python
> client = httpx.Client()
> client.get('https://example.org')
> ```

Seems tricky to implement, so let's rule it out-of-scope for this PR, and we can reconsider once we get the rest of #58  in.

flake8_trio.py
```diff
@@ -1441,9 +1467,131 @@ def visit_AsyncFunctionDef(
...
+            and isinstance(node.func, ast.Attribute)
+            and node.func.attr == "request"
+            and node.args
+            and isinstance(node.args[0], ast.Constant)
+            and isinstance(node.args[0].value, str)
+            and node.args[0].value.lower() in http_methods | {"trace", "connect"}
+        ) or (
+            "httpx" in self.imports
+            and isinstance(node.func, ast.Attribute)
+            and node.func.attr in http_methods | {"request", "stream", "send"}
```
Zac-HD, line 1536: I think this is probably too false-alarm-prone; without the extra evidence from the first argument being a known constant we'll error on *any* method with one of these extremely common names.  For example:

```python
import httpx

some_dict.get(...)  # should not raise a TRIO warning!
```

Zac-HD (commented):
I think we need to walk back one heuristic (see below); otherwise looks great - if you add the version bump we'll auto-release on merge!

flake8_trio.py
jakkdl: It still has to be inside an async function, and not be awaited. But yeah could also track the variable name a call to `httpx.Client()` is saved to (and do the same with urllib3), as is currently done with nurseries.

Removed it for now, and bumped version.

Zac-HD: > It still has to be inside an async function, and not be awaited. But yeah could also track the variable name a call to httpx.Client() is saved to (and do the same with urllib3), as is currently done with nurseries.

Probably a good follow-up PR once we get the other parts of #58 done 👍

Zac-HD (approved):

PR #98, opened 2023-01-12
Add TRIO900 - async generator without known safe decorator

### LINKED ISSUE #96: Error on async generator without known-safe decorators
[As discussed here](https://discuss.python.org/t/using-exceptiongroup-at-anthropic-experience-report/20888), async generators make it _very_ difficult to write cancellation-safe code, which is essential when you want to use timeouts (always) or directly cancel some tasks (not that rare either).  I therefore think it's worth having a disabled-by-default check, `TRIO900`, which bans *all* async generators unless they have an `@asynccontextmanager` decorator.

jakkdl:
Fixes #96 

Just over 30 minutes! Not a record, but :speedboat: 

Couldn't resist solving disabled-by-default with a decorator know that I'd figured out how to do that :innocent:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,12 @@
 # Changelog
 *[CalVer, YY.month.patch](https://calver.org/)*
 
+## 23.1.3
+- Add TRIO900: ban async generators not decorated with known safe decorator
+
 ## 23.1.2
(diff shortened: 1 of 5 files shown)


jakkdl:
rebased on top of main

PR #99, opened 2023-01-12
run all of pre-commit directly from the gh CI action

### LINKED ISSUE #97: Add pre-commit bot to repo
@Zac-HD With #88 merged you need to add https://pre-commit.ci/ to the repository for the checks to actually run. High priority, although I have the hook installed locally so won't push code that doesn't pass the checks.

jakkdl:
Fixes #97 
Re-renamed back typing to check, and GH CI now runs all of pre-commit and not just pyright.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -19,24 +19,6 @@ deps =
 commands =
     pytest {posargs} #{posargs:-n auto}
 
-[testenv:typing]
-description = Runs pyright in pre-commit from tox for CI, since it requires an internet connection
-deps =
-    pre-commit
-commands =
-    pre-commit run pyright --all-files
-
-# Settings for other tools
-[pytest]
-addopts =
-    --tb=native
-    --cov=flake8_trio
-    --cov-branch
-    --cov-report=term-missing:skip-covered
-    --cov-fail-under=100
-filterwarnings =
-    error
-
 
 [flake8]
 max-line-length = 90
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
     branches: [ main ]
 
 jobs:
-  typing:
+  check:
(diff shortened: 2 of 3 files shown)


jakkdl:
Looks like it has to reinitialize all the pre-commit environments on every run, whereas it'll reuse cached ones with the direct CI https://results.pre-commit.ci/run/github/514831223/1673533188.mO0ekGxrSQmivBCtgf0aCw but takes about the same amount of time.

Zac-HD:
> Looks like it has to reinitialize all the pre-commit environments on every run, whereas it'll reuse cached ones with the direct CI https://results.pre-commit.ci/run/github/514831223/1673533188.mO0ekGxrSQmivBCtgf0aCw but at least that run had to queue for a minute so it ended up taking about the same amount of time.

We can set up caching for Actions so that's fast too: https://tobiasmcnulty.com/posts/caching-pre-commit/

jakkdl:
Hrm, that means we need to start keeping around a `requirements.txt` file that needs to be updated every once in a while too. It's maybe possible to use setup.py, but regardless this starts getting kinda messy. Think I'm starting to prefer having the app added to the repo instead of manually messing around with files and caches and versions.

Zac-HD:
Fair point, let's not bother with caching then 😂

tox.ini
```diff
@@ -19,24 +19,6 @@ deps =
...
-[pytest]
-addopts =
-    --tb=native
-    --cov=flake8_trio
-    --cov-branch
-    --cov-report=term-missing:skip-covered
-    --cov-fail-under=100
-filterwarnings =
-    error
-
```
jakkdl, line 39: OOPS

jakkdl: we've been running without coverage since :sweat_smile:

PR #100, opened 2023-01-13
add trio240, usage of os.path in async function

jakkdl:
First bullet point of path manipulation in #58
Suspect this may be overly broad or trigger false alarms, but we'll see.

### DIFF
--- a/flake8_trio.py
+++ b/flake8_trio.py
@@ -25,7 +25,7 @@ if TYPE_CHECKING:
     from flake8.options.manager import OptionManager
 
 # CalVer: YY.month.patch, e.g. first release of July 2022 == "22.7.1"
-__version__ = "23.1.2"
+__version__ = "23.1.3"
 
 
 class Statement(NamedTuple):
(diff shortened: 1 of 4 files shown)


flake8_trio.py
```diff
@@ -1590,6 +1590,20 @@ def is_p_wait(arg: ast.expr) -> bool:
...
+class Visitor24X(Visitor200):
+    error_codes = {
+        "TRIO240": ("Avoid using os.path, prefer using trio.Path objects"),
+    }
+
+    def visit_Call(self, node: ast.Call):
+        if not self.async_function:
+            return
+        func_name = ast.unparse(node.func)
+        if func_name.startswith("os.path."):
```
Zac-HD, line 1603: Alas, I suspect that requiring rewrites for all code that calls `os.path.join()` (and other pure functions) is not going to be viable, so we'll need to read through https://docs.python.org/3/library/os.path.html and work out the list.

jakkdl: Read through the source code, and came up with this list
```python
    safe = (
        "_get_sep",
        "_varprog",
        "_varprogb",
        "normcase",
        "isabs",
        "join",
        "split",
        "splitext",
        "splitdrive",
        "basename",
        "dirname",
        "expanduser",
        "commonpath",
        "expandvars",  # os.fsencode, os.environ, os.fsdecode
        "abspath",  # calls, os.getcwd, which I think python tracks internally?
        "commonprefix",
        "samestat",  # just does simple equality check
        # constants
        "curdir",
        "pardir",
        "extsep",
        "sep",
        "pathsep",
        "defpath",
        "altsep",
        "devnull",
        "supports_unicode_filenames",
    )
    os_funcs = (
        "_path_normpath",
        "normpath",
        "_joinrealpath",
        "islink",
        "lexists",
        "ismount",  # safe on windows, unsafe on posix
        "realpath",
        "exists",
        "isdir",
        "isfile",
        "getatime",
        "getctime",
        "getmtime",
        "getsize",
        "samefile",
        "sameopenfile",
        "relpath",
    )
```

jakkdl:
Updated with explicit list of functions, also realized that looking for imports is easy - and might be somewhat common, so it will now catch that as well.
Rebased on top of main.

PR #101, opened 2023-01-13
Add TRIO23x, sync IO call in async function

jakkdl:
First bullet point of #58 
> Use `trio.open_file()` instead of `open()`, `io.open`, `io.open_code`, or `os.fdopen` in async functions

Although I guessed that `os.fdopen` should be wrapped in `trio.wrap_file` as per the intro

I have *no* idea how to stop shed/pyupgrade from reformatting `io.open` into `open`, tried both `#fmt: off` and `#noqa`.

### DIFF
--- a/tests/trio23x.py
+++ b/tests/trio23x.py
@@ -1,0 +1,42 @@
+# ARG --enable-visitor-codes-regex=(TRIO230)|(TRIO231)
+import io
+import os
+
+import trio
+
+
+async def foo():
+    open("")  # TRIO230: 4, 'open'
+    io.open_code("")  # TRIO230: 4, 'io.open_code'
+    os.fdopen(0)  # TRIO231: 4, 'os.fdopen'
+
+    # sync call is awaited, so it can't be sync
+    await open("")
+
+    # wrapped calls are always okay
+    await trio.wrap_file(open(""))
+    await trio.wrap_file(os.fdopen(0))
+
+    # with uses the same code & logic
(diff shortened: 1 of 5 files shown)


tests/trio23x.py
```diff
@@ -0,0 +1,42 @@
...
+    async with open(""):  # TRIO230: 15, 'open'
+        ...
+    async with trio.wrap_file(open("")):
+        ...
+
+    # test io.open
+    # pyupgrade removes the unnecessary `io.`
+    # https://github.com/asottile/pyupgrade#open-alias
+    # and afaict neither respects fmt:off nor #noqa - so I don't know how to test it
+    open("")  # TRIO230: 4, 'open'
```
Zac-HD, line 38: I think we just take it on faith that `io.open()` works too 😕 

(fwiw all `pyupgrade` or `shed` users will have the alias stripped anyway, so 🤷‍♂️)

Zac-HD (approved):
Looks good!

tests/trio23x.py
jakkdl: No yeah I'm not worried that would ever break, but a bit frustrating. Feature request for pyupgrade and/or shed?

Zac-HD: Shed would pass it through, `pyupgrade` would reject the feature request - asottile is (quite reasonably IMO) unwilling to add many/niche config options to the many key tools he maintains.

PR #102, opened 2023-01-16
Move eval files into a separate directory

jakkdl:
Taking the opportunity while there are no other open pull requests to move all the test_eval files to a separate directory. I've now also seen the light, and converted all paths from strings to Path. :lotus_position:


tests/test_decorator.py
```diff
@@ -80,7 +81,8 @@ def test_pep614():
     assert not wrap(("(any, expression, we, like)",), "no match here")
 
 
-common_flags = ["--select=TRIO", "tests/trio_options.py"]
+file_path = str(Path(__file__).parent / "trio_options.py")
+common_flags = ["--select=TRIO", file_path]
```
jakkdl, line 85: File path is now no longer hardcoded, but expects `trio_options.py` in the same directory. Not strictly necessary since this directory isn't moved, but makes it more robust.

tests/test_decorator.py
```diff
@@ -89,9 +91,9 @@ def test_command_line_1(capfd):
 
 
 expected_out = (
-    "tests/trio_options.py:2:1: TRIO107 "
+    f"{file_path}:5:1: TRIO107 "
     + Visitor107_108.error_codes["TRIO107"].format(
-        "exit", Statement("function definition", 2)
+        "exit", Statement("function definition", 5)
```
jakkdl, line 96: Update to using the new `file_path` variable instead of hardcoding the expected file location in the error message. See the comment for `trio_options.py` for the row change.

tests/trio_options.py
```diff
@@ -1,3 +1,6 @@
+app = None
+
+
 @app.route  # type: ignore
 async def f():
     print("hello world")
```
jakkdl, line 6: With these changes this file passes linters, so I don't have to exclude it in the rules. It was previously caught by the `tests/trio_*.py` pattern, but since it's not an eval'd file I didn't move it into the eval directory.
Although given the support for `#ARG` it's actually pretty simple to remove the dedicated test for this and make it an eval'd test file - but I'll leave that for a later TODO. Otherwise this file would probably benefit from a comment and/or another directory for `other_test_files` (or something)

tests/test_flake8_trio.py
```diff
@@ -25,13 +25,8 @@
...
-# TODO: Move test_eval files into a separate directory
-trio_test_files_regex = re.compile(r"trio\d.*.py")
-
-test_files: list[tuple[str, str]] = sorted(
-    (os.path.splitext(f)[0].upper(), f)
-    for f in os.listdir("tests")
-    if re.match(trio_test_files_regex, f)
+test_files: list[tuple[str, Path]] = sorted(
+    (f.stem.upper(), f) for f in (Path(__file__).parent / "eval_files").iterdir()
 )
```
jakkdl, line 30: No need for a regex, and using `Path` functions to do `os.path.splitext` (`.stem`) and `os.listdir` (`.iterdir()`).
It now also runs regardless of your current working directory, while previously it required being in the project root.

tests/test_flake8_trio.py
```diff
@@ -207,11 +202,6 @@ def test_noerror_on_sync_code(test: str, path: str):
     )
 
 
-def read_file(test_file: str):
-    filename = Path(__file__).absolute().parent / test_file
-    return Plugin.from_filename(str(filename))
-
-
```
jakkdl, line 214: With moving everything to using `Path` this function can now be completely removed :tada:

tests/test_flake8_trio.py
```diff
@@ -81,7 +76,7 @@ def test_eval(test: str, path: str):
 
     expected: list[Error] = []
 
-    with open(os.path.join("tests", path), encoding="utf-8") as file:
+    with open(path, encoding="utf-8") as file:
```
jakkdl, line 79: with `path` being a `Path` this can be simplified

tests/test_flake8_trio.py
```diff
@@ -146,7 +141,7 @@ def test_eval(test: str, path: str):
     assert parsed_args, "no parsed_args"
     assert expected, f"failed to parse any errors in file {path}"
 
-    plugin = read_file(path)
+    plugin = Plugin.from_filename(path)
```
jakkdl, line 144: With `Plugin.from_filename` taking a `Path`, and `path` containing the full path, we don't need the helper function at all here.

tests/test_flake8_trio.py
```diff
@@ -189,10 +184,10 @@ def visit_AsyncFor(self, node: ast.AST):
 
 
 @pytest.mark.parametrize(("test", "path"), test_files)
-def test_noerror_on_sync_code(test: str, path: str):
+def test_noerror_on_sync_code(test: str, path: Path):
     if any(e in test for e in error_codes_ignored_when_checking_transformed_sync_code):
         return
-    with tokenize.open(f"tests/{path}") as f:
+    with tokenize.open(path) as f:
```
jakkdl, line 190: More simplifying thanks to `Path`.

jakkdl (commented):
This is overall a pretty simple PR, with only ~30 lines changed and a bunch of files moved without any edits to them. I don't see much risk of unexpected consequences, since that should be caught by tests not running properly (which happened plenty while writing this :grin:), and there should be zero functionality change or anything noticeable for the end user of the plugin.

flake8_trio.py
```diff
@@ -1739,7 +1740,7 @@ def __init__(self, tree: ast.AST):
         self._tree = tree
 
     @classmethod
-    def from_filename(cls, filename: str) -> Plugin:
+    def from_filename(cls, filename: str | Path) -> Plugin:
```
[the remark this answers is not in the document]
Zac-HD: I think to be excruciatingly correct we could specify `os.PathLike` here, but I think the union is fine 😃

Zac-HD (approved):

jakkdl:
TIL about `os.PathLike`, will sneak the change into another PR ^^

PR #103, opened 2023-01-19
Split main file

jakkdl:
Kind of a massive overhaul, and somewhat WIP with some file names and structure being stuff I plan to improve on. But this should make working with the files, seeing diffs, interdependence between visitors/tests/files, and all that jazz be easier.

It's rebased on top of #102.


jakkdl:
oops, looks like I broke installation again. Will look at it later today

jakkdl:
Yay, about time I started figuring out how python packaging works.
Reasonably happy with this structure.

jakkdl:
Ran the code with `flake8-type-checking` to move some more imports into `TYPE_CHECKING` blocks since I'd already mostly done that anyway. It also wanted casts to be with string literals, but I ignored that.

Started going through the code to write a review summarizing changes, but since there's barely any code changes it was more like writing file docstrings ... so I just did that instead and enabled `flake8-docstrings` (but with most docstring requirements ignored). Will do a quick review of the code that's actually meaningful to look at, but otherwise it's just the file structure/organization that's up for review - don't bother looking for differences in the contained code unless I've marked it.

tests/test_flake8_trio.py
```diff
@@ -351,14 +356,15 @@ def info_tuple(error: Error):
 
     for err, exp in zip(errors, expected):
         err_msg = info_tuple(err)
-        for err, type_ in zip(err_msg, (int, int, str, type)):
+        for err, type_ in zip(err_msg, (int, int, str, type(None))):
```
jakkdl, line 359: This fourth parameter is a relic, and flake8 explicitly don't care about it anymore. During refactoring I had trouble returning the type of the plugin and started returning `type(None)` - which required this minor change in the tests.

tox.ini
```diff
@@ -22,7 +22,8 @@ commands =
 
 [flake8]
 max-line-length = 90
-extend-ignore = S101
+extend-ignore = S101, D101, D102, D103, D105, D106, D107, TC006
+per-file-ignores = flake8_trio/visitors/__init__.py: F401, E402
```
jakkdl, line 26: D101 to D107 is for flake8-docstrings for functions, classes, etc etc. If you want I could enable some of these and get to documenting stuff better.
TC006 is for `typing.cast` where the type isn't given as a string, I didn't see any big reason to bother with that so I ignored it.
`visitors/__init__.py` has the very non-standard imports to populate `ERROR_CLASSES` and `default_disabled_decorators` and noqa'ing each line was fugly, so added them here.

flake8_trio/base.py
```diff
@@ -0,0 +1,54 @@
...
+        self.code = error_code
+        self.message = message
+        self.args = args
+
+    # for yielding to flake8
+    def __iter__(self):
+        yield self.line
+        yield self.col
+        yield f"{self.code} " + self.message.format(*self.args)
+        yield None
```
jakkdl, line 39: This line (39) previous read `yield type(Plugin)`, but due to file moving/splitting the diff isn't picked up.

To avoid a circular import, and placing Plugin and error in the same file would also be problematic, I stop adhering to an outdated requirement that flake8 plugins should yield the type of their Plugin. I'm not finding it in their documentation now, but I remember the video tutorial mentioning that this parameter doesn't actually matter.

flake8_trio/visitors/visitors.py
```diff
@@ -0,0 +1,415 @@
...
+
+# used in 100
+checkpoint_node_types = (ast.Await, ast.AsyncFor, ast.AsyncWith)
+
+
+@error_class
+class Visitor100(Flake8TrioVisitor):
+    error_codes = {
+        "TRIO100": (
+            "{} context contains no checkpoints, add `await trio.lowlevel.checkpoint()`"
```
Zac-HD, line 26: "remove the context or add ..." (to suggest the cleaner minimalist fix)

Zac-HD, line 26: "remove the context or add ..." (to suggest the cleaner minimalist fix)

flake8_trio/visitors/visitors.py
```diff
@@ -0,0 +1,415 @@
...
+    "serve_listeners",
+    "serve",
+)
+
+
+@error_class
+class Visitor113(Flake8TrioVisitor):
+    error_codes = {
+        "TRIO113": (
+            "Dangerous `.start_soon()`, process might not run before `__aenter__` "
```
Zac-HD, line 258:
```suggestion
            "Dangerous `.start_soon()`, function might not be executed before `__aenter__` "
```

flake8_trio/visitors/visitors.py
```diff
@@ -0,0 +1,415 @@
...
+            "Startable function {} not in --startable-in-context-manager parameter "
+            "list, please add it so TRIO113 can catch errors using it."
+        ),
+    }
+
+    def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef):
+        if any(
+            isinstance(n, ast.arg) and n.arg == "task_status"
+            for n in self.walk(
+                *node.args.args, *node.args.posonlyargs, *node.args.kwonlyargs
```
Zac-HD, line 319: Pretty sure this *shouldn't* check posonlyargs actually, IIRC Trio always passes it by name.

flake8_trio/visitors/visitors.py
```diff
@@ -0,0 +1,415 @@
...
+    def visit_Call(self, node: ast.Call):
+        if get_matching_call(node, "sleep") and len(node.args) == 1:
+            arg = node.args[0]
+            if (
+                # `trio.sleep(math.inf)`
+                (
+                    isinstance(arg, ast.Attribute)
+                    and isinstance(arg.value, ast.Name)
+                    and arg.attr == "inf"
+                    and arg.value.id == "math"
```
Zac-HD, line 364: We might want to generalize this by dropping the `arg.value.id` check so that it catches e.g. `np.inf` too.

flake8_trio/visitors/visitors.py
```diff
@@ -0,0 +1,415 @@
...
+    def __init__(self, *args: Any, **kwargs: Any):
+        super().__init__(*args, **kwargs)
+        self.unsafe_function: ast.AsyncFunctionDef | None = None
+
+    def visit_AsyncFunctionDef(
+        self, node: ast.AsyncFunctionDef | ast.FunctionDef | ast.Lambda
+    ):
+        self.save_state(node, "unsafe_function")
+        if isinstance(node, ast.AsyncFunctionDef) and not fnmatch_qualified_name(
+            node.decorator_list, *self.options.no_checkpoint_warning_decorators
```
Zac-HD, line 405: I think here that we should *only* exempt `asynccontextmanager`; as an optional check it's appropriate to be stricter and I really do want to get rid of all the existing non-context-manager cases I know about - they generally make it impossible to handle exceptions correctly, which includes cancellation and therefore timeouts.

Zac-HD (commented):
OK, took me a while to read (and I imagine you a while to write!), but it all looks good to me.

Comments below can be addressed in a follow-up PR rather than this one, to minimize the review/merge-conflict/general overhead burden.  And thanks again!

tox.ini
jakkdl: I think enforcing per-file docstrings is good when we're starting to hit 20 files - and me documenting stuff should make reviewing easier too~

flake8_trio/__init__.py
```diff
@@ -0,0 +1,146 @@
+"""A highly opinionated flake8 plugin for Trio-related problems.
+
+This can include anything from outright bugs, to pointless/dead code,
+to likely performance issues, to minor points of idiom that might signal
+a misunderstanding.
+
+It may well be too noisy for anyone with different opinions, that's OK.
+
+Pairs well with flake8-async and flake8-bugbear.
```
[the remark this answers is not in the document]
jakkdl: Made a comment in #58 to follow up on this
https://github.com/Zac-HD/flake8-trio/issues/58#issuecomment-1406352603

tests/test_decorator.py
```diff
@@ -90,10 +94,17 @@ def test_command_line_1(capfd):
     assert capfd.readouterr() == ("", "")
 
 
+expected_lineno = -1
+with open(file_path) as f:
+    for lineno, line in enumerate(f):
+        if line.startswith("async def"):
+            expected_lineno = lineno + 1
```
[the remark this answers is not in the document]
jakkdl: ooh, me like!

setup.py
```diff
@@ -18,12 +19,12 @@ def local_file(name: str) -> Path:
         raise AssertionError("__version__ must be set")
 
 
-setuptools.setup(
+setup(
     name="flake8-trio",
     version=__version__,
     author="Zac Hatfield-Dodds and Contributors",
```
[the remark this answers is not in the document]
jakkdl: :heart: 
I don't think we've had any other contributors, so could remove that unless we want to leave it in for if/when it happens.

flake8_trio/visitors/visitors.py
jakkdl: Yep, the doc confirms
> The async_fn must accept a `task_status` keyword argument, [...]

https://trio.readthedocs.io/en/stable/reference-core.html#trio.Nursery.start

Zac-HD:
Haha, yeah, I trusted that was the case but GitHub doesn't have a good moved-code view so I just read the lot 😅

flake8_trio/visitors/visitors.py
jakkdl: Addressing this in a separate PR.

setup.py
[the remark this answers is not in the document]
Zac-HD: Nobody else has contributed code to this repo, but I do think of the people in the Trio issue threads and Gitter who suggested lint rules as contributors to the project 😃... and I'd prefer to be early than late in welcoming more people to contribute.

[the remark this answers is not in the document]
jakkdl: Makes sense!

PR #104, opened 2023-01-19
Add TRIO232: blocking sync call on file objects

jakkdl:
A first draft at tracking types for file/path/httpx/etc and warning on sync calls. Except this one just does it for `open()` and the varios `io.*` types it can return as a start.
This and others that build on top of it should finish off #58 

Branched from on top of #103 - but you can just view the latest commit to see the contents of this one.

* The logic for tracking variables and their type will be moved to a superclass that different error_classes can inherit from (or possibly just implemented in `Flake8TrioVisitor`.
  * Child classes will probably be able to register calls and their return type so the superclass can use that. (so something like `super().register('open', 'io.TextIOWrapper')` in the `__init__`
  * code can then maybe be reused for the error classes tracking nurseries
* I don't think anybody in their right mind is going to send around `io.TextIOWrapper` objects to functions, but that's going to be much more common with Path and httpx objects, so I implemented it for this as well just to show the logic.
* Unless reimplemented as a plugin an top of a type checker this will ofc never be perfect, but handling `Optional[Path]` and `Path|None` will be done. `list[Path]` and similar cases can be done if it makes sense but I don't think will be common of any of the types we care about.
* I thiiink scope rules are handled correctly at the moment, but if that turns out to be tricky it can probably be dropped without worry.
* Code in general isn't super clean and will be rewritten, so don't bother with nitpicking much/at all

### DIFF
--- a/tests/eval_files/trio232.py
+++ b/tests/eval_files/trio232.py
@@ -1,0 +1,135 @@
+import io
+from io import BufferedRandom, BufferedReader, BufferedWriter, TextIOWrapper
+from typing import Any, Optional
+
+blah: Any = None
+
+
+async def file_text(f: io.TextIOWrapper):
+    f.read()  # TRIO232: 4, 'read', 'f'
+
+
+async def file_binary_read(f: io.BufferedReader):
+    f.read()  # TRIO232: 4, 'read', 'f'
+
+
+async def file_binary_write(f: io.BufferedWriter):
+    f.read()  # TRIO232: 4, 'read', 'f'
+
+
+async def file_binary_readwrite(f: io.BufferedRandom):
+    f.read()  # TRIO232: 4, 'read', 'f'
+
+
+async def file_text_(f: TextIOWrapper):
+    f.read()  # TRIO232: 4, 'read', 'f'
+
+
+async def file_binary_read_(f: BufferedReader):
+    f.read()  # TRIO232: 4, 'read', 'f'
+
+
+async def file_binary_write_(f: BufferedWriter):
+    f.read()  # TRIO232: 4, 'read', 'f'
+
+
+async def file_binary_readwrite_(f: BufferedRandom):
+    f.read()  # TRIO232: 4, 'read', 'f'
+
+
+async def file_text_3(f: TextIOWrapper = blah):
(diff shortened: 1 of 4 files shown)


Zac-HD:
Approach seems promising, agree that `list[whatever]` isn't worth handling but `whatever|None` probably is (if easy!).  I look forward to seeing where this goes after rebasing 😁

jakkdl:
Rebased on top of main (which wasn't trivial since it was forked from an outdated #103 that was later amended - fun stuff!).
Will now continue working on it~

jakkdl:
I'll refactor it as I write the next object sync checker, as this one should work well as a standalone.

tests/eval_files/trio232.py
```diff
@@ -0,0 +1,135 @@
...
+    f.read()  # TRIO232: 4, 'read', 'f'
+
+
+async def file_text_3(f: TextIOWrapper = blah):
+    f.read()  # TRIO232: 4, 'read', 'f'
+
+
+async def file_text_4(f: TextIOWrapper | None):
+    if f:
+        f.read()  # TRIO232: 8, 'read', 'f'
```
Zac-HD, line 46: Would this still give a lint error without the `if`, or in an `else`?  It'd be nice to demonstrate that with tests.

Zac-HD (approved):
Tiny comment on tests, otherwise looks great.

Zac-HD:
For the future refactored version, one annoying pattern is defining a `httpx.Client()` as a global variable...

tests/eval_files/trio232.py
jakkdl: Yes, it currently does not care at all that it could be a None - but I had type-checking turned on in my editor when writing this one hence the `if`'s. You're completely right it should be shown in tests though - and will fix in followup PR :+1:

jakkdl:
> For the future refactored version, one annoying (to lint, it's fortunately rare) pattern is defining a `httpx.Client()` as a global variable...

oh and it should work fine with global variables as long as they're defined in the same file. Will add tests to show that.


I'll make a PR just with these two fixes + another pattern I just realized will be useful:

```python
async def type_assign():
    f: TextIOWrapper = ...
    f.read()
```

PR #107, opened 2023-01-27
Small followup fixes and changes to 114 & 116 after the file split

jakkdl:
Addressing comments in #103 

Only question mark is whether to also remove "and Contributors" in `setup.py`.

Also added an extra newline in a test, cuz it was bothering me with the way I run tests.


Zac-HD (approved):
Sweet!

PR #108, opened 2023-01-27
TRIO900 now only checks for asynccontextmanager, ignoring flags

jakkdl:
Split this out cause I wasn't sure if it was going to be a bigger PR - but ended up being quite manageable.


Zac-HD (approved):
Nice - I'll merge this and #107 locally so we can ship them in a single release 👍

PR #109, opened 2023-01-27
Fix TRIO103 false alarm on excepts after a previous except has caught Cancelled

### LINKED ISSUE #106: False alarm for `TRIO103`, on `except BaseException` after an `except Cancelled` clause
```python
try:
    ...
except trio.Cancelled:
    raise
except BaseException:
    pass  # this should be OK, because the error _can't_ be a Cancelled.
```

We should also make the README and the message for BaseException suggest this pattern.

jakkdl:
Fixes #106 

Also fixes several other cases not mentioned in the issue, and does the same for TRIO104.

Took me a minute to understand the existing 103/104 implementation and where to inject the logic for this, but ultimately a pretty simple addition.

It broke, like, *all* of the 103/104 tests though since they didn't bother having separate `try`s for the excepts, so went ham on adding those. But I didn't bother fixing `trio103_no_104.py` and cut that down to the minimal tests needed to check what that one actually checks. (that 103 is enabled, and not 104, when only one of them is selected with the regex).
But at the end of `trio103.py` and `trio104.py` are the new test cases added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## Future
 - Add TRIO232: blocking sync call on file object.
 
+## Future
+- TRIO103 and TRIO104 no longer triggers when `trio.Cancelled` has been handled in previous except handlers.
+
 ## 23.1.4
 - TRIO114 no longer triggers on posonly args named "task_status"
 - TRIO116 will now match on any attribute parameter named `.inf`, not just `math.inf`.
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ pip install flake8-trio
 - **TRIO101**: `yield` inside a nursery or cancel scope is only safe when implementing a context manager - otherwise, it breaks exception handling.
 - **TRIO102**: it's unsafe to await inside `finally:` or `except BaseException/trio.Cancelled` unless you use a shielded
   cancel scope with a timeout.
-- **TRIO103**: `except BaseException` and `except trio.Cancelled` with a code path that doesn't re-raise.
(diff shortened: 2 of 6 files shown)


jakkdl:
> We should also make the README and the message for BaseException suggest this pattern.

If i try to get python highlighting in the code block in the readme, shed errors out
```
Could not parse ```python markdown block in file 'README.md'
    InvalidInput: Cannot parse: 1:0: except trio.Cancelled:(
```
since an `except` without a `try` is invalid syntax. Not sure if it's possible to tell shed (or whatever part of shed it is that errors out) to ignore the block.

Not sure how to format it the neatest inside the bullet list anyway.


Having different messages for an error hasn't actually been done before, and doing it this way is *quite* clunky. It feels like there should be some way of specifying the error message as like
```
"TRIO103": "{0} block with a code path that doesn't re-raise the error.{'more text here' if {0} != 'trio.Cancelled else ''}",
```
and get `str.format` to be as powerful as an f-string, but some very quick googling only turns up eval'ing it with an `f` added to the beginning of it.
Not too necessary for this case, but if we do it more (and the 2xx's would probably have used this instead of having separate codes - which I didn't wanna do in this case) I'd like to get a better solution.

I also just realized that I've been horribly inconsistent in whether error messages end with a period or not.

jakkdl:
With `pass` and `raise` looking somewhat similar when viewing the diff, I went ahead and replaced all the `pass` with `...` in the test files so it's more clear where there's a `raise`.

Zac-HD (commented):
Adjusting error message will need some test changes, otherwise looks great!

jakkdl:
Committed suggestions, fixed tests, and squashed.

Zac-HD (approved):

PR #112, opened 2023-01-28
add TRIO117, error on any reference to multierror

### LINKED ISSUE #105: Do not raise or catch `trio.[NonBase]MultiError`; prefer `[exceptiongroup.]BaseExceptionGroup`.
Trio's `MultiError` is deprecated in favor of the builtin (or backport of) `ExceptionGroup`, so downstream code should be written or ported to the latter.  That includes catching, raising, and use of methods - so I think just linting for any reference to `trio.MultiError` would be sufficient and easy to implement.

This lint is important because `MultiError` is still raised from Trio, in order to give downstream time to migrate, and so users quite naturally use the name mentioned in their tracebacks.

jakkdl:
Fixes #105 
I made some assumptions when writing the README message, and the error message could maybe be more informative.
Otherwise a very simple and straightforward check. I could restrict it to raise/catch/methods if this is too liberal, but simplest was to just trigger on any name/attribute and not care about context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 
 ## Future
 - TRIO103 and TRIO104 no longer triggers when `trio.Cancelled` has been handled in previous except handlers.
-- Add TRIO117: Reference to `[NonBase]MultiError`
+- Add TRIO117: Reference to deprecated `trio.[NonBase]MultiError`; use `[Base]ExceptionGroup` instead.
 
 ## 23.1.4
 - TRIO114 no longer triggers on posonly args named "task_status"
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 
 ## Future
 - TRIO103 and TRIO104 no longer triggers when `trio.Cancelled` has been handled in previous except handlers.
+- Add TRIO117: Reference to `[NonBase]MultiError`
```
Zac-HD, line 8:
```suggestion
- Add TRIO117: Reference to deprecated `trio.[NonBase]MultiError`; use `[Base]ExceptionGroup` instead.
```

Zac-HD (approved):

PR #113, opened 2023-01-30
Fix typos

kianmeng:
Found via `codespell -S .mypy_cache -L aline`


jakkdl:
Thank you! I'll look into adding codespell to the pre-commit config :)

jakkdl:
I suspect that `aline` is a typo too, so won't be excluding that one in the config.

kianmeng:
:partying_face: :partying_face: :partying_face: :partying_face: :partying_face:

PR #118, opened 2023-01-31
add TRIO212: sync call on httpx.Client, move type-tracking logic

jakkdl:
Ticks another box in #58 

All the code ... just worked .. on the first try?? I got so paranoid about the testing infrastructure being broken, but nope - turns out I was just writing correct code over and over.

Not the prettiest class structure since the type-tracking class can't be reused for classes that don't want to inherit from Visitor200, but I'll fix that if/when I want to do that. Or does nurseries want this functionality? Or anything else?

The diff looks relatively large because a big chunk of code is moved without being modified, but this should otherwise be pretty straightforward to review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 *[CalVer, YY.month.patch](https://calver.org/)*
 ## Future
 - Add TRIO232: blocking sync call on file object.
+- Add TRIO212: blocking sync call on httpx.Client object.
 
 ## Future
 - TRIO103 and TRIO104 no longer triggers when `trio.Cancelled` has been handled in previous except handlers.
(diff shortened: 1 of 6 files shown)


jakkdl:
CI found a type in 232, that I think got missed by tests not being rerun on the TRIO232 PR after codespell was added to pre-commit (and it wasn't merged when codespell was run manually in #113). So that's also randomly fixed

jakkdl:
Also added urllib3 pools. Didn't add explicit blocking methods on those since they're so many, but presumably there might be some. (Though probably easier to add a whitelist of allowed method calls).
Not the prettiest code, but should be fine enough.

Zac-HD (approved):
🚀

Zac-HD:
Nice!  I'll aim to release in the next few days.

PR #120, opened 2023-02-02
Add support for the anyio library

### LINKED ISSUE #61: Support checking `anyio` code too
[`anyio` is a compatibility layer on top of `asyncio` or Trio](https://anyio.readthedocs.io/en/stable/), allowing you to support trio-like [structured concurrency](https://en.wikipedia.org/wiki/Structured_concurrency) using either.  In practice, this means that most Trio-supporting libraries have a free way to support asyncio too, and I expect it'll be more generally helpful for the asyncio ecosystem as `TaskGroup`s become more popular over the coming years.

Since almost all of `flake8-trio`'s checks are actually about structured concurrency rather than being truly library-specific, and `anyio` matches the `trio` API quite closely, it would be relatively easy to add support for `anyio`.  There are a couple of small divergences, but I think it would mostly be a matter of checking twice as many names.

jakkdl:
Fixes #61
Have fun with this tiny commit :upside_down_face: 

Kind of a chore, I maybe should have gone at some parts in different ways, but the library should now have full support for anyio - and all eval files are tested with "trio" replaced with "anyio", looking to see that all the errors are still good, and that there's no instances of "trio" in the output.

If you trigger an error without importing either trio or anyio, the library will default to saying "trio" in error messages. If you trigger an error having imported both, it will say stuff like
> sync call ... in async function, use `[trio|anyio].open_file(...)`.

Otherwise it's not too much that's actually changed, just lots of small replacements in a billion places.

### DIFF
--- a/flake8_trio/visitors/flake8triovisitor.py
+++ b/flake8_trio/visitors/flake8triovisitor.py
@@ -28,6 +28,16 @@ class Flake8TrioVisitor(ast.NodeVisitor):
         self.outer: dict[ast.AST, dict[str, Any]] = {}
         self.novisit = False
 
+        # check whether library we're working towards has been explicitly
+        # specified with --anyio, otherwise assume Trio - but we update if we
+        # see imports
+        if self.options.anyio:
+            self.library_explicit = True
+            self.library = "anyio"
+        else:
+            self.library_explicit = False
+            self.library = "trio"
+
     def visit(self, node: ast.AST):
         """Visit a node."""
         # construct visitor for this node type
@@ -117,3 +127,17 @@ class Flake8TrioVisitor(ast.NodeVisitor):
     def walk(self, *body: ast.AST) -> Iterable[ast.AST]:
         for b in body:
             yield from ast.walk(b)
+
+    def visit_Import(self, node: ast.Import):
+        for alias in node.names:
+            name = alias.name
+            if name in ("trio", "anyio") and alias.asname is None:
+                # if a library has not been explicitly imported or specified,
+                # replace the default
+                if not self.library_explicit:
+                    self.library = name
+                # otherwise, append this to the messages - if it's not already in it
+                elif name not in self.library:
+                    self.library = f"[{self.library}|{name}]"
+
+                self.library_explicit = True
--- a/flake8_trio/visitors/helpers.py
+++ b/flake8_trio/visitors/helpers.py
@@ -136,15 +136,16 @@ cancel_scope_names = (
 
 
 # convenience function used in a lot of visitors
+# should probably return a named tuple
 def get_matching_call(
-    node: ast.AST, *names: str, base: str = "trio"
-) -> tuple[ast.Call, str] | None:
+    node: ast.AST, *names: str, base: Iterable[str] = ("trio", "anyio")
+) -> tuple[ast.Call, str, str] | None:
     if (
         isinstance(node, ast.Call)
         and isinstance(node.func, ast.Attribute)
         and isinstance(node.func.value, ast.Name)
-        and node.func.value.id == base
+        and node.func.value.id in base
         and node.func.attr in names
     ):
-        return node, node.func.attr
+        return node, node.func.attr, node.func.value.id
     return None
--- a/flake8_trio/visitors/visitors.py
+++ b/flake8_trio/visitors/visitors.py
@@ -22,8 +22,8 @@ checkpoint_node_types = (ast.Await, ast.AsyncFor, ast.AsyncWith)
 class Visitor100(Flake8TrioVisitor):
     error_codes = {
         "TRIO100": (
-            "{} context contains no checkpoints, remove the context or add"
-            " `await trio.lowlevel.checkpoint()`"
+            "{0}.{1} context contains no checkpoints, remove the context or add"
+            " `await {0}.lowlevel.checkpoint()`"
         ),
     }
 
@@ -34,7 +34,8 @@ class Visitor100(Flake8TrioVisitor):
                 isinstance(x, checkpoint_node_types) and x != node
                 for x in ast.walk(node)
             ):
-                self.error(item, f"trio.{call[1]}")
+                # type checking has been done inside get_matching_call
+                self.error(item, call[2], call[1])
 
     visit_AsyncWith = visit_With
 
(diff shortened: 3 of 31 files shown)


flake8_trio/visitors/visitor103_104.py
```diff
@@ -70,10 +71,11 @@ def visit_ExceptHandler(self, node: ast.ExceptHandler):
...
         if self.unraised and marker is not None:
             extra_message = ""
-            if marker.name != "trio.Cancelled":
+            if not re.fullmatch(r"(trio|anyio)\.Cancelled", marker.name):
+                library = self.library if self.library else "trio"
                 extra_message = (
-                    " Consider adding an `except trio.Cancelled: raise` before this"
-                    " exception handler."
+                    f" Consider adding an `except {library}.Cancelled: raise`"
+                    " before this exception handler."
```
Zac-HD, line 78: See `Cancelled` comment above.

flake8_trio/visitors/visitors.py
```diff
@@ -127,35 +128,38 @@ def visit_Call(self, node: ast.Call):
             and isinstance(node.func, ast.Attribute)
             and isinstance(node.func.value, ast.Name)
             and (
-                (node.func.value.id == "trio" and node.func.attr in trio_async_funcs)
+                (
+                    node.func.value.id in ("trio", "anyio")
+                    and node.func.attr in trio_async_funcs
```
Zac-HD, line 133: Some of `trio_async_funcs` don't exist in `anyio`, and vice-versa; we should probably make a second list (well, tuple) for `anyio` and check them separately.

flake8_trio/visitors/visitors.py
```diff
@@ -386,12 +394,16 @@ class Visitor117(Flake8TrioVisitor):
...
 
     def visit_Attribute(self, node: ast.Attribute):
-        if (n := ast.unparse(node)) in ("trio.MultiError", "trio.NonBaseMultiError"):
-            self.error(node, n)
+        if (
+            isinstance(node.value, ast.Name)
+            and node.value.id in ("trio", "anyio")
+            and node.attr in DEPRECATED_ERRORS
+        ):
+            self.error(node, ast.unparse(node))
```
Zac-HD, line 406: `anyio` doesn't have the old `MultiError`, just `ExceptionGroup` 🙂

flake8_trio/visitors/flake8triovisitor.py
```diff
@@ -117,3 +118,12 @@ def save_state(self, node: ast.AST, *attrs: str, copy: bool = False):
...
             yield from ast.walk(b)
+
+    def visit_Import(self, node: ast.Import):
+        for alias in node.names:
+            name = alias.name
+            if name in ("trio", "anyio") and alias.asname is None:
+                if self.library is not None and name != self.library:
+                    self.library = f"[{self.library}|{name}]"
+                else:
+                    self.library = alias.name
```
Zac-HD, line 129: This is great, but I think we should also have a global `--anyio` flag to set the default library; otherwise `anyio` projects will have `trio` recommended to them in any file that doesn't happen to import `anyio`!

flake8_trio/visitors/helpers.py
```diff
@@ -103,10 +103,10 @@ def has_exception(node: ast.expr | None) -> str:
         if (
             isinstance(node, ast.Attribute)
             and isinstance(node.value, ast.Name)
-            and node.value.id == "trio"
+            and node.value.id in ("trio", "anyio")
             and node.attr == "Cancelled"
         ):
-            return "trio.Cancelled"
+            return f"{node.value.id}.Cancelled"
```
Zac-HD, line 109: Unfortunately `anyio` has a *slightly* different API here; see [`anyio.get_cancelled_exc_class()`](https://anyio.readthedocs.io/en/stable/api.html#anyio.get_cancelled_exc_class).  That's important because you can have both `trio` and `asyncio` in the same program (ugh, but sometimes the least-bad option!).

We probably also want to lint for "no assigning that result because it breaks our linter, as well as multi-backend programs".

Zac-HD (commented):
This is *shockingly* easy for such a conversion - huge kudos to you for an implementation that made this feasible, and @agronholm for a super-clean drop in library 🙌

A few things before this is ready to ship are listed in comments below; happy to do those here or in a follow-up PR as you prefer.

🤔 I should probably also run this past @agronholm too, in case I missed something...

flake8_trio/visitors/visitors.py
jakkdl: oh hum, it also doesn't have `.open_nursery()`

flake8_trio/visitors/helpers.py
jakkdl: I assume you mean `anyio` and not `asyncio`? Maybe not the best name they could've chosen :sweat_smile: 

I'll probably push this to a separate issue+PR.

jakkdl: :scream:

jakkdl: Tracked in #122

jakkdl:
Fixed all review comments (or disabled functionality and moved out to separate issues). Also did some refactoring to make stuff cleaner, added `# NOTRIO` and `# ANYIO_NO_ERROR` (those magic markers should perhaps have some checks that they're not misspelled or something, or use some other way of being more robust)

tests/eval_files/trio105_anyio.py
```diff
@@ -0,0 +1,13 @@
+# NOTRIO
+import anyio
+
+
+async def foo():
+    nursery = anyio.open_nursery()
```
Zac-HD, line 6: Probably worth a comment that all this is fine because anyio calls the equivalent construct a "taskgroup", following asyncio?

tests/eval_files/trio102_anyio.py
```diff
@@ -0,0 +1,28 @@
+import anyio
+
+
+async def foo4():
+    try:
+        ...
+    except ValueError:
+        await foo()  # safe
+    except anyio.Cancelled:
```
Zac-HD, line 9: I presume this will be updated in a future PR 👍

Zac-HD (approved):

tests/eval_files/trio105_anyio.py
agronholm: Following curio, actually. Asyncio got task groups much later than either curio or AnyIO.

tests/eval_files/trio102_anyio.py
agronholm: Yeah, there is no `anyio.Cancelled`.

agronholm: The AnyIO way to do this is `except anyio.get_cancelled_exc_class():`.

Zac-HD: Yep, discussed in https://github.com/Zac-HD/flake8-trio/pull/120#discussion_r1095001688 and tracked in #122 🙂

jakkdl: Yeah that check is to make sure it doesn't raise an error - regardless of if it exists as a library function or not. Can add a comment with #122

PR #126, opened 2023-02-06
reenable coverage, fix missing coverage

jakkdl:
only introduced one branch not covered by tests during the three weeks with disabled coverage at least ^^

Accidentally removed in #99

### DIFF
--- a/flake8_trio/__init__.py
+++ b/flake8_trio/__init__.py
@@ -43,7 +43,8 @@ class Plugin:
         self._tree = tree
 
     @classmethod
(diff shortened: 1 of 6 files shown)


jakkdl:
Getting it to work post-filesplit without `--devel` was surprisingly tricky.
I found https://github.com/pytest-dev/pytest-cov/issues/98 and needed to both add a `tests/__init__.py` and add a `[coverage:paths]` section.

jakkdl:
I suppose this is a marginal reason to run coverage outside of GH tasks and have a fancy badge in the readme, hah

PR #127, opened 2023-02-06
Add support for anyio.get_cancelled_exc_class() in TRIO102, 103 and 104.

### LINKED ISSUE #106: False alarm for `TRIO103`, on `except BaseException` after an `except Cancelled` clause

### LINKED ISSUE #122: Handle critical exceptions in anyio
Matters w/r/t TRIO102, 103 and 104.

flake8-trio currently always just looks for `trio.Cancelled`, but
https://github.com/Zac-HD/flake8-trio/pull/120#discussion_r1095001688
> Unfortunately `anyio` has a _slightly_ different API here; see [`anyio.get_cancelled_exc_class()`](https://anyio.readthedocs.io/en/stable/api.html#anyio.get_cancelled_exc_class). That's important because you can have both `trio` and `asyncio` in the same program (ugh, but sometimes the least-bad option!).

Also add another error code for assigning the result of `anyio.get_cancelled_exc_class()` to a variable. That can be caught if within the same file with our rudimentary type tracking, but not if done across files and imported. So the plugin should likely track assignments for cases where single-backend is used and the error code gets disabled. 
> We probably also want to lint for "no assigning that result because it breaks our linter, as well as multi-backend programs".

jakkdl:
Fixes #122 

rewrote `critical_except` a bit as I updated it, in large part because I personally failed to read the old version of the code :innocent: 
Fix #106 for TRIO102 (#109 only fixed TRIO103 & TRIO104)
Thanks to #125 it was quite smooth to add a different suggestion for TRIO103 when anyio is the primary library.
As I started to write a `assert_correct_message` test helper I realized I had pretty bad code duplication, so checking of column, message and args are now all handled in `assert_correct_attribute`.
Improved how `ignore_column` is handled in anyio tests.
A slew of tests being moved around and modified and stuff so the necessary eval_files are handled with trio/anyio as appropriate. And some additional ones for #106+TRIO102


jakkdl:
Added a separate error message when both are available.

But I'm unsure how to handle TRIO103 when both are imported, should bare excepts require that both have been handled in that case?

flake8_trio/visitors/visitor103_104.py
```diff
@@ -16,19 +16,27 @@
...
 _trio103_common_msg = "{} block with a code path that doesn't re-raise the error."
+_suggestion = " Consider adding an `except {}: raise` before this exception handler."
+_suggestion_dict = {
+    "trio": "trio.Cancelled",
+    "anyio": "anyio.get_cancelled_exc_class()",
+}
+_suggestion_dict[
+    "[trio|anyio]"
+] = f'[{_suggestion_dict["trio"]}|{_suggestion_dict["anyio"]}]'
+_suggestion_dict["[anyio|trio]"] = _suggestion_dict["[trio|anyio]"]
```
Zac-HD, line 27: Let's canonicalize this, so we only have to worry about one ordering.

I'd also recommend a tuple literal rather than the (not-valid-at-runtime) `|`.

tests/eval_files/trio102.py
```diff
@@ -162,6 +162,55 @@ async def foo3():
...
 
+async def foo4():
+    try:
+        ...
+    except ValueError:
+        await foo()  # safe
+    except BaseException:
+        await foo()  # error: 8, Statement("BaseException", lineno-1)
+    except:
+        await foo()  # safe, since BaseException will catch Cancelled
```
Zac-HD, line 173: ...safe, yes, but also redundant with the `except BaseException:` 🤔

Zac-HD (approved):
This looks good to me!

Some non-blocking nitpicks below; I intend to try it out on some `anyio` libraries to check that the results looks reasonable, and then merge if there's no obvious problems.

Given that LibCST support seems close-ish, if it's noisy I'll wait for auto-fixes before recommending to OSS maintainers.

Zac-HD:
> I'm unsure how to handle TRIO103 when both are imported, should bare excepts require that both have been handled in that case?

I think in this case it's safe to just catch `anyio.get_cancelled_exc_class()`, because if you're actually running Trio code that will return `trio.Cancelled`.

tests/eval_files/trio102.py
jakkdl: I didn't actually know that exceptions *have* to inherit from BaseException, interesting. I opened https://github.com/PyCQA/flake8-bugbear/issues/349 in case anybody cares for it.

jakkdl:
> > I'm unsure how to handle TRIO103 when both are imported, should bare excepts require that both have been handled in that case?
> 
> I think in this case it's safe to just catch `anyio.get_cancelled_exc_class()`, because if you're actually running Trio code that will return `trio.Cancelled`.

If you're running all three async libraries, and anyio's current backend is `asyncio`, and in the `try:` statement stuff is ran that could either raise `trio.Cancelled` or `anyio.get_cancelled_exc_class()`, it might bork? But that seems weird enough that I think not raising false alarms is more important.

flake8_trio/visitors/visitor103_104.py
jakkdl: I fiddled a bit with this and ... maybe the code is better now? Idk, I'll likely fiddle around with this more when I write more library-specific/dependent code

jakkdl:
addressed nitpicks and rebased on top of main to ensure coverage.

PR #130, opened 2023-02-08
Fix get_matching_call when passed a single string as base.

### LINKED ISSUE #129: False-alarm in `TRIO210` - dicts, queues, etc are not sync http clients
```python
async def f():
    r = {}
    r.get("not a sync http client")  # False alarm: raises TRIO210
```

I've seen similar issues on `Queue().put(...)`, `SomeCls.delete()`, etc.

jakkdl:
Fixes #129 
Introduced in 73c9fc50 when adding support for anyio.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Changelog
 *[CalVer, YY.month.patch](https://calver.org/)*
+## Future
+- Fix get_matching_call when passed a single string as base. Resolves possibly several false alarms, TRIO210 among them.
 
 ## 23.2.2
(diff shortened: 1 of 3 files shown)


jakkdl:
I don't think this is the first time that I've been bit by `assert_type("", Iterable[str])`. Feels like something that should either be warned upon by type checkers ("don't pass a string to a function expecting an `Iterable[str]`") - or maybe just a check for bugbear to never use `Iterable[str]`.

Zac-HD (approved):
Probably not worth linting for this, it's a footgun but also too useful to ban 😕 

PR looks great though, thanks 🙏

jakkdl:
Hm, might bring it up with mypy/pyright or discuss.python.org - but that's for another time.

Issue #121, opened 2023-02-02
Implement TRIO105 for anyio (library function must be awaited)


jakkdl:
From https://github.com/Zac-HD/flake8-trio/pull/120#discussion_r1094980781

> Some of trio_async_funcs don't exist in anyio, and vice-versa; we should probably make a second list (well, tuple) for anyio and check them separately.

I can probably compile it without too much issue, but let's have a separate issue & PR for it.

jakkdl:
nursery calls have also been implemented in this since #56, that should probably be rewritten to use the new type-tracking functionality (Do we also want name-based tracking on top of that?), and anyio's version of that is Task Groups and https://anyio.readthedocs.io/en/stable/api.html#anyio.create_task_group

Zac-HD:
> nursery calls have also been implemented in this since https://github.com/Zac-HD/flake8-trio/issues/56, that should probably be rewritten to use the new type-tracking functionality (Do we also want name-based tracking on top of that?)

Skimming through that issue, we only used name-based tracking at all because we thought type-based tracking wasn't worth the trouble.  Since we have the latter now let's add that; but keep name-based checks for `nursery` because there are a few patterns where you pass that around as an argument.

For `TaskGroup`, it looks like the conventional names are covered by `tg|task_?group`.

jakkdl:
> because there are a few patterns where you pass that around as an argument.

This only matters if they're not typing the parameter list, if they do it's covered.

While giving false positives on non-nursery objects named `tg` isn't too bad since it's an awful name, it almost surely will give false positives.

Zac-HD:
Hmm, we [currently](https://github.com/Zac-HD/flake8-trio/blob/cffff4a1826e542ef24d1b3753fe31dfbd1dd2bf/flake8_trio/visitors/visitors.py#L107-L134) only match `.*nursery\.start`; I think `(tg|.*(nursery|task_?group)\.start` wouldn't have too many false alarms?  Important that we're looking at a method call not just the variable name though.

jakkdl:
Ah yes, I was assuming I'll import over the type-tracking functionality used in 2xx

jakkdl:
[...]

PR #135, opened 2023-02-13
Add anyio.TaskGroup support to TRIO113

jakkdl:
Stemming from #121 

rewriting nursery detection to use type tracking - affecting TRIO105&TRIO113

Much of the diff is due to renaming `eval_files/trio113.py` to `eval_files/trio113_trio.py` with very minor changes, and having checks that aren't trio-specific in `eval_files/trio113.py`

I've realized that typing the eval files is often quite handy, and also to check pyright/mypy support for whatever check I'm writing. So I modified the config to longer blanket exclude eval_files, and added `type: ignore` at the top of all files except the ones I directly worked on in this one. Might gradually remove some of the other ones too.
Likewise got a couple `# mypy: disable-error-code` since I (still) am running that in my editor. It raises a lot more warnings than pyright, but most of the code in the repo passes it (without `strict`)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,6 +22,8 @@ repos:
           - hypothesmith
           - pytest
           - flake8
+          - trio
+          - anyio
 
 -   repo: https://github.com/codespell-project/codespell
     rev: v2.2.2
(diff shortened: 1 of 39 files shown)


Zac-HD (approved):
🤩

jakkdl:
Haha, the perfect review comment :grin:

PR #139, opened 2023-02-14
Fix pyright config since file structure changes, enable optional checks, fix type errors

jakkdl:
I've been bouncing some issues at https://github.com/microsoft/pyright and during that realized that the `strict` setting hadn't been updated since the file structure was changed - oopsies.
Also tested *all* optional checks that aren't turned on, even with `strict`, and other than unused `type: ignore` which collides massively with mypy, I left them all on :innocent: 
And ofc fixed all the warnings everywhere as a result of the ones above. The missing super() calls in classes that don't inherit is a bit silly, but otherwise they all seem reasonable and did catch some potentially useful things in the tests.


flake8_trio/__init__.py
```diff
@@ -155,7 +156,7 @@ def parse_trio200_dict(raw_value: str) -> dict[str, str]:
             # if we raise it as ValueError
             raise ArgumentTypeError(
                 f"Invalid number ({len(split_values)-1}) of splitter "
-                f"tokens {splitter!r} in {value!r}",
+                + f"tokens {splitter!r} in {value!r}"
```
Zac-HD, line 159: Implicit concatenation should be fine here?

Zac-HD (approved):

PR #140, opened 2023-02-14
False alarm on fixture decorator

### LINKED ISSUE #137: `@pytest.fixture()` should disable `TRIO101`
As for `TRIO900` in 3f3c13c9db8caadcca4d9ad7ca1df44d4db81805.

jakkdl:
Forked from on top of #139 to have correct type checking.

fixes #137 - and also added fixture to 91X

Harmonized decorator checking so 101, 113, 900, 910 and 911 now all use `has_decorator` - which itself uses `_get_identifier`. And made them all (except for 113, which I'm not sure about) check for `pytest.fixture[()]`

It's a bit messy that all the different checks specify different lists of ignored decorators. Feels like that should be a constant somewhere - but the lists *are* slightly different, though I'm not sure they need to be.

91X: overload, fixture - and `no-checkpoint-warning-decorators` defaults to asynccontextmanager
101: contextmanager, asynccontextmanager, fixture
113: asynccontextmanager
900: asynccontextmanager, fixture

overload can clearly be added to all of them without any issue, I'm not sure about 113+fixture, and 101 needs contextmanager, but I don't know if the others care about it.

The remaining raw checks against node.decorator_list are
91x: for additional `options.no_checkpoint_warning_decorators`, but that uses fnmatch.
visitor109: which completely bails out on any decorator


Zac-HD (approved):
💪🚢

PR #142, opened 2023-02-19
reimplement trio100 & trio101 with libcst

jakkdl:
Also changed how trio100 works to avoid missed cases where a function is defined inside the with statement.
100 & 101 moved into their own files.

I spent some time trying to get batching to work, `BatchedVisitor` is not compatible with `MatcherDecoratableVisitor` - and especially not with transformers. Tried dynamically creating a class that inherited from all enabled visitors, but ran into so many minor headaches with that approach I abandoned it. Might revisit once all visitors are changed to cst, but for now it's certainly not worth the effort for the sake of speeding it up. There might be a case of wanting to do the transforms simultaneously in case one transform unlocks another, and stuff like that.
I thought batching would be required, since the utility visitors are separate classes, but I think that can be made to work by having error classes that require the functionality of a utility visitor to inherit from it. But neither 100 nor 101 require any of the utility ones so didn't bother with it too much.

I'm waiting with enabling transforming for trio100 for a later PR, since that will require a bunch of UI changes / flags / etc.

Ended up opening a bunch of issues in pyright/libcst/mypy during the process, https://github.com/Instagram/LibCST/issues/870 is probably the most relevant one - which is probably gonna be somewhat of a pain going forward if not fixed. (I didn't bother trying to stop shed from removing `Union`, but if possible that'd make it less painful.)


jakkdl:
Ah, looks like the CI failure is due to shed not fixing everything in one go. I've noticed that happening frequently which is somewhat annoying, will track down the cause and open an issue in shed.

jakkdl:
Fixed the CI, and cribbed your try/finally solution for `LIBCST_PARSER_TYPE` - except switched out `pop()` for `del` to avoid a `reportUnusedCallResult`

flake8_trio/visitors/visitor100.py
```diff
@@ -0,0 +1,67 @@
...
+    def leave_With(self, original_node: cst.With, updated_node: cst.With) -> cst.With:
+        if not self.has_checkpoint_stack.pop():
+            res = self.node_dict[original_node]
+            self.error(original_node, *res)
+        # if: autofixing is enabled for this code
+        # then: remove the with and pop out it's body
+        return updated_node
+
+    @m.visit(m.Await() | m.For(asynchronous=m.Asynchronous()))
+    # can't use m.call_if_inside(m.With), since it matches parents *or* the node itself
```
Zac-HD, line 53: We might want to write a helper function for this at some point?

tests/eval_files/trio100.py
```diff
@@ -35,17 +44,35 @@ async def function_name():
...
+        pass
+
 
 async def function_name2():
-    with (
+    # now throws the error at the with statement
+    with (  # error: 4, "trio", "fail_after"
         open("") as _,
-        trio.fail_after(10),  # error: 8, "trio", "fail_after"
+        trio.fail_after(10),
```
Zac-HD, line 68: Hmm, I prefer the more precise (old) locations - can we get that back, maybe [via `WithItem`](https://libcst.readthedocs.io/en/latest/nodes.html#libcst.WithItem)?

Zac-HD (commented):
This is looking nice!  I think I was anticipating a considerably gnarlier migration...

I'd appreciate a response to `super().__init__()` and the `WithItem` thing; other comments don't need to be addressed before merging or perhaps at all.

flake8_trio/visitors/helpers.py
```diff
@@ -174,3 +189,49 @@ def get_matching_call(
...
+    node: cst.With, *names: str, base: Iterable[str] = ("trio", "anyio")
+) -> tuple[cst.Call, str, str] | None:
+    for item in node.items:
+        if res := m.extract(
+            item.item,
+            m.Call(
+                func=m.Attribute(
+                    value=m.SaveMatchedNode(m.Name(), name="library"),
+                    attr=m.SaveMatchedNode(
+                        m.OneOf(*(m.Name(x) for x in names)),
```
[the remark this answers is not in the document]
jakkdl: ooh and map is cleaner too. cribbed ^^

flake8_trio/visitors/helpers.py
```diff
@@ -174,3 +189,49 @@ def get_matching_call(
...
+                    ),
+                )
+            ),
+        ):
+            assert isinstance(item.item, cst.Call)
+            assert isinstance(res["library"], cst.Name)
+            assert isinstance(res["function"], cst.Name)
+            if res["library"].value not in base:
+                return None
+            return item.item, res["function"].value, res["library"].value
```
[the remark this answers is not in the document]
jakkdl: Yeah it's quite overdue (the ast version, `get_matching_call` has had a comment saying it should be done for quite some time), and now is as good a time as any :+1:

flake8_trio/visitors/visitor100.py
jakkdl: Oooh, I hadn't even considered writing my own decorators. Hell yes I want this, would be very useful in shed as well - I'm pretty sure there's at least one bad matcher due to this. But yeah, for another PR.

tests/eval_files/trio100.py
jakkdl: Fixed! ^^

flake8_trio/visitors/helpers.py
```diff
@@ -174,3 +189,49 @@ def get_matching_call(
...
+    return None
+
+
+def func_has_decorator(func: cst.FunctionDef, *names: str) -> bool:
+    namenode = m.Name(value=m.MatchIfTrue(lambda x: x in names))
+
+    return m.matches(
+        func,
+        m.FunctionDef(
```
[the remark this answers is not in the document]
jakkdl: pulled out a `list_contins` - it's close to being usable in `with_has_item` as well, will likely iterate on these as I use them in more rewritten visitors/transformers.

jakkdl:
> This is looking nice! I think I was anticipating a considerably gnarlier migration...
you keep underestimating how good the infrastructure I've built up is :grin: Though I did spend a lot of time on polishing this, and there's quite a bit of potential future gnarl left.

flake8_trio/visitors/helpers.py
```diff
@@ -174,3 +189,63 @@ def get_matching_call(
...
+                )
+            ),
+        ):
+            assert isinstance(item.item, cst.Call)
+            assert isinstance(res["library"], cst.Name)
+            assert isinstance(res["function"], cst.Name)
+            if res["library"].value not in base:
+                return None
+            return AttributeCall(item.item, res["library"].value, res["function"].value)
+    return None
```
Zac-HD, line 236: I think here we actually want to continue looping if it's an unknown base?  More generally I have a (weak) preference to surface all the errors we can find in a single shot, and so I'd be inclined to return a `list[AttributeCall]`.

flake8_trio/visitors/visitor100.py
```diff
@@ -0,0 +1,67 @@
...
+        self.has_checkpoint_stack: list[bool] = []
+        self.node_dict: dict[cst.With, AttributeCall] = {}
+
+    def visit_With(self, node: cst.With) -> None:
+        if m.matches(node, m.With(asynchronous=m.Asynchronous())):
+            self.checkpoint_node(node)
+        if res := with_has_call(
+            node, "fail_after", "fail_at", "move_on_after", "move_on_at", "CancelScope"
+        ):
+            self.node_dict[node] = res
```
Zac-HD, line 38: Here, if we return each match, I think we can `for r in res: self.node_dict[r.node] = r`.  I think that restores reporting each distinct warning if there are multiple calls in a single `with`-statement?

Zac-HD:
> you keep underestimating how good the [infra]structure/patterns I've built up are 😁

[Hofstadter's law](https://en.wikipedia.org/wiki/Hofstadter%27s_law) has earned my respect, but you're right - kudos!

Zac-HD (commented):
Tiny tweak below, but this looks good to me.  I'm inclined to merge (with or without tweak) in the morning and release a new version, what do you reckon?

flake8_trio/visitors/helpers.py
jakkdl: Oops yes, it should very much continue looping.

flake8_trio/visitors/visitor100.py
jakkdl: leave_with would also have to be changed to loop over it's items (and `leave_WithItem` is not an option since we need to parse it after the body has been parsed). Changing the dict to contain a sequence is probably the way to go.
I went with the lazy option cause I didn't think multiple withitems was used much ever, but I'll fix it in a yiffie~

jakkdl:
Tweaked and ready to be shipped! And good call, this is nicer.

flake8_trio/visitors/helpers.py
jakkdl: and yep, changed it to return a list.

Zac-HD (approved):
🚢

flake8_trio/visitors/visitor100.py
```diff
@@ -0,0 +1,67 @@
...
+        # if: autofixing is enabled for this code
+        # then: remove the with and pop out it's body
+        return updated_node
+
+    @m.visit(m.Await() | m.For(asynchronous=m.Asynchronous()))
+    # can't use m.call_if_inside(m.With), since it matches parents *or* the node itself
+    # need to use Union due to https://github.com/Instagram/LibCST/issues/870
+    def checkpoint_node(self, node: Union[cst.Await, cst.For, cst.With]):
+        if self.has_checkpoint_stack:
+            self.has_checkpoint_stack[-1] = True
```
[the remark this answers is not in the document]
Zac-HD: See https://discuss.python.org/t/pep-649-deferred-evaluation-of-annotations-tentatively-accepted/21331 for background, this is a long-running problem which should at least get better from 3.12 😌

PR #143, opened 2023-02-24
reimplement 91X in libcst

jakkdl:
This one is *chonky*
I might go over it tomorrow and add review comments to help reviewing it. Although there's not really much to do other than grind through it - and running it on existing code, I wouldn't be shocked if there's crashes or bugs that didn't exist before, so maybe do that before releasing.

I generally optimized for minimizing the diff, rather than writing the best / most logical code - might leave that for a second commit, or do it when implementing autofix. Esp any `isinstance`s should generally not be there when working against libcst.

I rewrote the loop checker to not require two passes, by injecting a fake statement and doing some fancy logic. Quite happy I came up with this solution, much better than the previous super messy one.
Removing redundant checkpoints might be more doable with this approach as well.
Having to access metadata to get the linenos of nodes is quite ugly, with the redundant checkpoint I'll be looking to store nodes instead of `Statement`s in `uncheckpointed_statements`

test_910_permutations is now *extremely* slow, previously being so fast I never noticed it, but now it takes *50 seconds* on my machine. I did some rudimentary timing, and the `parse_module` call took 5 of the seconds, and plugin.run() taking 45 seconds. Now that is across ~4000 calls, so nothing too crazy, but I'll probably put it behind `--runfuzz` if I don't find anything responsible for the slowdown. 91X currently doesn't use any decorator markers, so changing it to be a simple transformer might do wonders.

A few changes in eval_files:
1. realized that comprehensions aren't handled, added regression tests.
2. I think the nested error test ... was wrong? I don't see why that should give two errors per line.
3. multiple yields in a single boolop is no longer handled as before. I maybe should've just done `fmt: off` instead of fighting over and over with black/shed

### DIFF
--- a/flake8_trio/visitors/flake8triovisitor.py
+++ b/flake8_trio/visitors/flake8triovisitor.py
@@ -8,7 +8,6 @@ from abc import ABC
 from typing import TYPE_CHECKING, Any, Union
 
 import libcst as cst
-import libcst.matchers as m
 from libcst.metadata import PositionProvider
 
 from ..base import Error, Statement
@@ -46,8 +45,6 @@ class Flake8TrioVisitor(ast.NodeVisitor, ABC):
             "outer",
             "typed_calls",
         }
-
-        self.suppress_errors = False
 
     # `variables` can be saved/loaded, but need a setter to not clear the reference
     @property
@@ -105,17 +102,16 @@ class Flake8TrioVisitor(ast.NodeVisitor, ABC):
         elif not re.match(self.options.enable_visitor_codes_regex, error_code):
             return
 
-        if not self.suppress_errors:
-            self.__state.problems.append(
-                Error(
-                    # 7 == len('TRIO...'), so alt messages raise the original code
-                    error_code[:7],
-                    node.lineno,
-                    node.col_offset,
-                    self.error_codes[error_code],
-                    *args,
-                )
+        self.__state.problems.append(
+            Error(
+                # 7 == len('TRIO...'), so alt messages raise the original code
+                error_code[:7],
+                node.lineno,
+                node.col_offset,
+                self.error_codes[error_code],
+                *args,
             )
+        )
 
     def get_state(self, *attrs: str, copy: bool = False) -> dict[str, Any]:
         if not attrs:
@@ -131,8 +127,6 @@ class Flake8TrioVisitor(ast.NodeVisitor, ABC):
 
     def set_state(self, attrs: dict[str, Any], copy: bool = False):
         for attr, value in attrs.items():
-            if copy and hasattr(value, "copy"):
-                value = value.copy()
             setattr(self, attr, value)
 
     def save_state(self, node: ast.AST, *attrs: str, copy: bool = False):
@@ -163,14 +157,14 @@ class Flake8TrioVisitor(ast.NodeVisitor, ABC):
             self.__state.library = self.__state.library + (name,)
 
 
-class Flake8TrioVisitor_cst(m.MatcherDecoratableTransformer, ABC):
+class Flake8TrioVisitor_cst(cst.CSTTransformer, ABC):
     # abstract attribute by not providing a value
     error_codes: dict[str, str]  # pyright: reportUninitializedInstanceVariable=false
     METADATA_DEPENDENCIES = (PositionProvider,)
 
     def __init__(self, shared_state: SharedState):
         super().__init__()
-        self.outer: dict[cst.BaseStatement, dict[str, Any]] = {}
+        self.outer: dict[cst.CSTNode, dict[str, Any]] = {}
         self.__state = shared_state
 
         self.options = self.__state.options
@@ -193,7 +187,7 @@ class Flake8TrioVisitor_cst(m.MatcherDecoratableTransformer, ABC):
                 value = value.copy()
             setattr(self, attr, value)
 
-    def save_state(self, node: cst.BaseStatement, *attrs: str, copy: bool = False):
+    def save_state(self, node: cst.CSTNode, *attrs: str, copy: bool = False):
         state = self.get_state(*attrs, copy=copy)
         if node in self.outer:
             # not currently used, and not gonna bother adding dedicated test
@@ -201,6 +195,9 @@ class Flake8TrioVisitor_cst(m.MatcherDecoratableTransformer, ABC):
             self.outer[node].update(state)  # pragma: no cover
         else:
             self.outer[node] = state
+
+    def restore_state(self, node: cst.CSTNode):
+        self.set_state(self.outer.pop(node, {}))
 
     def error(
         self,
--- a/flake8_trio/visitors/helpers.py
+++ b/flake8_trio/visitors/helpers.py
@@ -7,10 +7,11 @@ from __future__ import annotations
 
 import ast
 from fnmatch import fnmatch
-from typing import TYPE_CHECKING, NamedTuple, TypeVar
+from typing import TYPE_CHECKING, Any, NamedTuple, TypeVar
 
 import libcst as cst
 import libcst.matchers as m
+from libcst.helpers import ensure_type, get_full_name_for_node_or_raise
 
 from ..base import Statement
 from . import (
@@ -27,6 +28,7 @@ if TYPE_CHECKING:
 
     T = TypeVar("T", bound=Flake8TrioVisitor)
     T_CST = TypeVar("T_CST", bound=Flake8TrioVisitor_cst)
+    T_EITHER = TypeVar("T_EITHER", Flake8TrioVisitor, Flake8TrioVisitor_cst)
 
 
 def error_class(error_class: type[T]) -> type[T]:
@@ -41,7 +43,7 @@ def error_class_cst(error_class: type[T_CST]) -> type[T_CST]:
     return error_class
 
 
-def disabled_by_default(error_class: type[T]) -> type[T]:
+def disabled_by_default(error_class: type[T_EITHER]) -> type[T_EITHER]:
     assert error_class.error_codes
     default_disabled_error_codes.extend(error_class.error_codes)
     return error_class
@@ -86,6 +88,19 @@ def fnmatch_qualified_name(name_list: list[ast.expr], *patterns: str) -> str | N
     return None
 
 
+def fnmatch_qualified_name_cst(
+    name_list: Iterable[cst.Decorator], *patterns: str
+) -> str | None:
+    for name in name_list:
+        qualified_name = get_full_name_for_node_or_raise(name)
+
+        for pattern in patterns:
+            # strip leading "@"s for when we're working with decorators
+            if fnmatch(qualified_name, pattern.lstrip("@")):
+                return pattern
+    return None
+
+
 # used in 103/104 and 910/911
 def iter_guaranteed_once(iterable: ast.expr) -> bool:
     # static container with an "elts" attribute
@@ -112,6 +127,8 @@ def iter_guaranteed_once(iterable: ast.expr) -> bool:
                     return True
             else:
                 return True
+        return False
+
     # check for range() with literal parameters
     if (
         isinstance(iterable, ast.Call)
@@ -120,6 +137,62 @@ def iter_guaranteed_once(iterable: ast.expr) -> bool:
     ):
         try:
             return len(range(*[ast.literal_eval(a) for a in iterable.args])) > 0
+        except Exception:  # noqa: PIE786
+            return False
+    return False
+
+
+def cst_literal_eval(node: cst.BaseExpression) -> Any:
+    ast_node = cst.Module([cst.SimpleStatementLine([cst.Expr(node)])]).code
+    try:
+        return ast.literal_eval(ast_node)
+    except Exception:  # noqa: PIE786
+        return None
+
+
+def iter_guaranteed_once_cst(iterable: cst.BaseExpression) -> bool:
+    # static container with an "elts" attribute
+    if elts := getattr(iterable, "elements", []):
+        for elt in elts:
+            assert isinstance(
+                elt,
+                (
+                    cst.Element,
+                    cst.DictElement,
+                    cst.StarredElement,
+                    cst.StarredDictElement,
+                ),
+            )
+            # recurse starred expression
+            if isinstance(elt, (cst.StarredElement, cst.StarredDictElement)):
+                if iter_guaranteed_once_cst(elt.value):
+                    return True
+            else:
+                return True
+        return False
+
+    if isinstance(iterable, cst.SimpleString):
+        return len(ast.literal_eval(iterable.value)) > 0
+
+    # check for range() with literal parameters
+    if m.matches(
+        iterable,
+        m.Call(
+            func=m.Name("range"),
+        ),
+    ):
+        try:
+            return (
+                len(
+                    range(
+                        *[
+                            cst_literal_eval(a.value)
+                            for a in ensure_type(iterable, cst.Call).args
+                        ]
+                    )
+                )
+                > 0
+            )
         except Exception:  # noqa: PIE786
             return False
     return False
--- a/flake8_trio/visitors/visitor100.py
+++ b/flake8_trio/visitors/visitor100.py
@@ -5,8 +5,9 @@ context does not contain any `await` statements.  This makes it pointless, as
 the timeout can only be triggered by a checkpoint.
 Checkpoints on Await, Async For and Async With
 """
-# if future annotations are imported then shed will reformat away the Union use
-from typing import Any, Union
+from __future__ import annotations
+
+from typing import Any
 
 import libcst as cst
 import libcst.matchers as m
@@ -29,9 +30,13 @@ class Visitor100_libcst(Flake8TrioVisitor_cst):
         self.has_checkpoint_stack: list[bool] = []
         self.node_dict: dict[cst.With, list[AttributeCall]] = {}
 
+    def checkpoint(self) -> None:
+        if self.has_checkpoint_stack:
+            self.has_checkpoint_stack[-1] = True
+
     def visit_With(self, node: cst.With) -> None:
         if m.matches(node, m.With(asynchronous=m.Asynchronous())):
-            self.checkpoint_node(node)
+            self.checkpoint()
         if res := with_has_call(
             node, "fail_after", "fail_at", "move_on_after", "move_on_at", "CancelScope"
         ):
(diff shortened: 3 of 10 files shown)


flake8_trio/visitors/visitor91x.py
```diff
@@ -329,37 +496,34 @@ def visit_loop(self, node: ast.While | ast.For | ast.AsyncFor):
             self.uncheckpointed_statements.update(self.uncheckpointed_before_break)
 
         # reset break & continue in case of nested loops
-        self.set_state(outer)
+        self.outer[node]["uncheckpointed_statements"] = self.uncheckpointed_statements
+        self.restore_state(node)
+        ...
```
jakkdl, line 501: surprised no flake8 plugin dislikes this random ellipsis

jakkdl:
Oh, I didn't check coverage before pushing. Will look at that another day.

jakkdl:
> test_910_permutations is now extremely slow, previously being so fast I never noticed it, but now it takes 50 seconds on my machine. I did some rudimentary timing, and the parse_module call took 5 of the seconds, and plugin.run() taking 45 seconds. Now that is across ~4000 calls, so nothing too crazy, but I'll probably put it behind --runfuzz if I don't find anything responsible for the slowdown. 91X currently doesn't use any decorator markers, so changing it to be a simple transformer might do wonders.

Switched to using CSTTransformer instead of m.MatcherDecoratableTransformer and that slashed the runtime of 910_permutations from 41.82s to 14.35s - so I'll default to CSTTransformer unless the decorators are actually useful. (Sidenote: I looked into writing my own, and it's not that easy). 14s is still enough to be kind of a hassle though, so added it to fuzz only.
The existing fuzz tests takes ... literal forever now though. I had it run for 17 minutes just now before ast.parse crashed on unicode/escape sequence crap.

jakkdl:
I had to add a bunch of stuff to eval files for coverage, since multiple visitors using the same helper shared the burden of covering them - but now that some of them are converted to libcst all the cases have to be covered both by cst visitors and ast visitors.

flake8_trio/visitors/flake8triovisitor.py
```diff
@@ -47,8 +46,6 @@ def __init__(self, shared_state: SharedState):
             "typed_calls",
         }
 
-        self.suppress_errors = False
-
```
jakkdl, line 51: no longer used since 91X doesn't visit loop bodies twice, where it previously needed to suppress errors during one of the passes to not duplicate errors.

flake8_trio/visitors/flake8triovisitor.py
```diff
@@ -131,8 +127,6 @@ def get_state(self, *attrs: str, copy: bool = False) -> dict[str, Any]:
 
     def set_state(self, attrs: dict[str, Any], copy: bool = False):
         for attr, value in attrs.items():
-            if copy and hasattr(value, "copy"):
-                value = value.copy()
```
jakkdl, line 135: was only used in 91x, could warrant a no-cover but /shrug

flake8_trio/visitors/flake8triovisitor.py
```diff
@@ -202,6 +196,9 @@ def save_state(self, node: cst.BaseStatement, *attrs: str, copy: bool = False):
         else:
             self.outer[node] = state
 
+    def restore_state(self, node: cst.CSTNode):
+        self.set_state(self.outer.pop(node, {}))
+
```
jakkdl, line 201: visitor classes are still responsible for restoring the outer state, which is not the case for ast visitors. Might change that at some point, but for now at least added a helper function.

flake8_trio/visitors/flake8triovisitor.py
```diff
@@ -163,14 +157,14 @@ def add_library(self, name: str) -> None:
             self.__state.library = self.__state.library + (name,)
 
 
-class Flake8TrioVisitor_cst(m.MatcherDecoratableTransformer, ABC):
+class Flake8TrioVisitor_cst(cst.CSTTransformer, ABC):
```
jakkdl, line 160: ~3x speedup, and wasn't really using decorators anyway

flake8_trio/visitors/helpers.py
```diff
@@ -112,6 +127,8 @@ def iter_guaranteed_once(iterable: ast.expr) -> bool:
                     return True
             else:
                 return True
+        return False
+
```
jakkdl, line 131: mostly just saves on doing another isinstance call on line 134, no logic change, but noticed when I was debugging coverage.

flake8_trio/visitors/helpers.py
```diff
@@ -125,6 +142,62 @@ def iter_guaranteed_once(iterable: ast.expr) -> bool:
     return False
 
 
+def cst_literal_eval(node: cst.BaseExpression) -> Any:
+    ast_node = cst.Module([cst.SimpleStatementLine([cst.Expr(node)])]).code
+    try:
+        return ast.literal_eval(ast_node)
+    except Exception:  # noqa: PIE786
+        return None
```
jakkdl, line 150: major hoops to have to run through, but didn't find any better way of doing it for now.

flake8_trio/visitors/helpers.py
```diff
@@ -125,6 +142,62 @@ def iter_guaranteed_once(iterable: ast.expr) -> bool:
...
+
+
+def iter_guaranteed_once_cst(iterable: cst.BaseExpression) -> bool:
```
jakkdl, line 153: ugly, but works pretty much like the ast version.

flake8_trio/visitors/visitor100.py
```diff
@@ -5,8 +5,9 @@
 the timeout can only be triggered by a checkpoint.
 Checkpoints on Await, Async For and Async With
 """
-# if future annotations are imported then shed will reformat away the Union use
-from typing import Any, Union
+from __future__ import annotations
+
+from typing import Any
```
jakkdl, line 10: not using decorators -> don't have to care about libcst's type checking not handling `|`

flake8_trio/visitors/visitor91x.py
```diff
@@ -7,37 +7,89 @@
...
+                        ]
+                    )
+                )
+            ]
+        ),
+    )
+
+
+@dataclass
+class LoopState:
```
jakkdl, line 52: So many state variables I decided to toss them into dataclasses. It mostly just makes save_state/restore_state stuff easier and less prone to missing a variable.

flake8_trio/visitors/visitor91x.py
```diff
@@ -54,148 +106,232 @@ def __init__(self, *args: Any, **kwargs: Any):
...
-
         self.uncheckpointed_statements: set[Statement] = set()
-        self.uncheckpointed_before_continue: set[Statement] = set()
-        self.uncheckpointed_before_break: set[Statement] = set()
 
-        self.default = self.get_state()
+        self.loop_state = LoopState()
+        self.try_state = TryState()
+
+    @property
```
jakkdl, line 114: I elected to add a million property and setters instead of rewriting everything refer to `self.loop_state.xxx`. In large part to make the diff smaller / easier to read, but I don't care much for them otherwise.

flake8_trio/visitors/visitor91x.py
```diff
@@ -54,148 +106,232 @@ def __init__(self, *args: Any, **kwargs: Any):
...
+                node.decorators, *self.options.no_checkpoint_warning_decorators
+            )
         )
         if not self.async_function:
             return
 
+        pos = self.get_metadata(PositionProvider, node).start
         self.uncheckpointed_statements = {
-            Statement("function definition", node.lineno, node.col_offset)
+            Statement("function definition", pos.line, pos.column)
```
jakkdl, line 192: will quite possibly change this in future PR's to save nodes instead of statements, just gotta figure out how that plays with tests.

flake8_trio/visitors/visitor91x.py
```diff
@@ -54,148 +106,232 @@ def __init__(self, *args: Any, **kwargs: Any):
...
 
-    # disregard checkpoints in nested function definitions
-    def visit_FunctionDef(self, node: ast.FunctionDef):
-        self.save_state(node)
-        self.set_state(self.default, copy=True)
+        return updated_node
 
-    # checkpoint functions
-    def visit_Await(self, node: ast.Await | ast.Raise):
+    def error_91x(
```
jakkdl, line 220: added error helper with the introduction of artificial statements/errors - which is the new solution to not knowing the state of the beginning of loops before they've been fully parsed.

flake8_trio/visitors/visitor91x.py
```diff
@@ -54,148 +106,232 @@ def __init__(self, *args: Any, **kwargs: Any):
...
+        return updated_node
 
     # valid checkpoint if there's valid checkpoints (or raise) in:
     # (try or else) and all excepts, or in finally
     #
     # try can jump into any except or into the finally* at any point during it's
     # execution so we need to make sure except & finally can handle worst-case
     # * unless there's a bare except / except BaseException - not implemented.
-    def visit_Try(self, node: ast.Try):
+    def visit_Try(self, node: cst.Try):
```
jakkdl, line 282: visit_Try was previously a massive function, it's now split into several different attribute visitor and leavers - but retains pretty much the exact same logic.

flake8_trio/visitors/visitor91x.py
```diff
@@ -204,123 +340,154 @@ def visit_Try(self, node: ast.Try):
...
 
-            self.visit_nodes(node.finalbody)
-            self.uncheckpointed_statements.difference_update(added)
+    def leave_Try(self, original_node: cst.Try, updated_node: cst.Try) -> cst.Try:
+        self.restore_state(original_node)
+        return updated_node
 
-    # valid checkpoint if both body and orelse checkpoint
-    def visit_If(self, node: ast.If | ast.IfExp):
+    def leave_If_test(self, node: cst.If | cst.IfExp) -> None:
```
jakkdl, line 356: visit_If is also split up into multiple functions

flake8_trio/visitors/visitor91x.py
```diff
@@ -204,123 +340,154 @@ def visit_Try(self, node: ast.Try):
...
+        if self.async_function:
+            # merge current state with post-body state
+            self.uncheckpointed_statements.update(
+                self.outer[original_node]["uncheckpointed_statements"]
+            )
+        return updated_node
+
+    # libcst calls attributes in the order they appear in the code, so we manually
+    # rejig the order here
+    def visit_IfExp(self, node: cst.IfExp) -> bool:
```
jakkdl, line 384: could in ast be handled the exact same way, but not requires some slight rejigging

flake8_trio/visitors/visitor91x.py
```diff
@@ -204,123 +340,154 @@ def visit_Try(self, node: ast.Try):
...
+            copy=True,
+        )
+        self.loop_state = LoopState()
+        self.infinite_loop = self.body_guaranteed_once = False
 
     # Check for yields w/o checkpoint in between due to entering loop body the first time,
     # after completing all of loop body, and after any continues.
     # yield in else have same requirement
     # state after the loop same as above, and in addition the state at any break
-    def visit_loop(self, node: ast.While | ast.For | ast.AsyncFor):
```
jakkdl, line 241: visit_loop now also split into a lot of different visit/leavers, it retains a lot of logic but changes the logic as previously mentioned from doing two passes of the loop body (first one to calculate what state the loop can be in when it starts, the second one to actually raise errors depending on that state), to injecting a special statement into `uncheckpointed_statements` that makes errors raised become "artificial errors", and depending on what the potential end-state of the loop is we either toss those out or turn them into actual errors.

flake8_trio/visitors/visitor91x.py
```diff
@@ -329,37 +496,33 @@ def visit_loop(self, node: ast.While | ast.For | ast.AsyncFor):
...
+    def visit_Break(self, node: cst.Break):
+        self.has_break = True
         if not self.async_function:
             return
         self.uncheckpointed_before_break.update(self.uncheckpointed_statements)
 
     # first node in a condition is always evaluated, but may shortcut at any point
     # after that so we track worst-case checkpoint (i.e. after yield)
-    def visit_BoolOp(self, node: ast.BoolOp):
+    def visit_BooleanOperation_right(self, node: cst.BooleanOperation):
```
jakkdl, line 518: minor logic diff, ast will save `x and y and z` as a single `And` with three values, but cst always parses boolops as binary trees. This leads to slight differences in behaviour as well - see eval files

tests/eval_files/trio100.py
```diff
@@ -40,10 +40,15 @@ async def function_name():
...
             pass
+
     async with trio.fail_after(10):
         async for _ in receive_channel:
             pass
 
+    async with trio.fail_after(10):  # error: 15, "trio", "fail_after"
+        for _ in receive_channel:
+            pass
+
```
jakkdl, line 51: codecov, non-async for does not checkpoint.

tests/eval_files/trio103.py
```diff
@@ -226,6 +226,54 @@ def foo() -> Any:
             continue
         raise
 
+try:
+    ...
+except BaseException:  # TRIO103_trio: 7, "BaseException"
+    for i in foo():
+        raise
+
```
jakkdl, line 234: codecov for AST `guaranteed_iter_once`

tests/eval_files/trio910.py
```diff
@@ -518,3 +529,12 @@ async def foo_test_return():
...
     return
+
+
+async def foo_comprehension_1():
+    [... for x in range(10) if await foo()]
+
+
+# should error
+async def foo_comprehension_2():
+    [await foo() for x in range(10) if foo()]
```
jakkdl, line 540: comprehensions are not handled at all, so any `await` inside them will be treated as always being executed. They also weren't handled previously

tests/eval_files/trio910.py
```diff
@@ -510,6 +511,16 @@ async def foo_test2():
...
+@custom_disabled_decorator
+async def foo_ARG_test():
+    print("...")
+
+
+@custom_disabled_decorator()
+async def foo_ARG_test2():
+    print("...")
+
+
```
jakkdl, line 523: idr if this actually covered more, but no reason not to have a test for it in the eval file

tests/eval_files/trio911.py
```diff
@@ -161,8 +161,8 @@ async def foo_while_5():
     while foo():
         yield  # error: 8, "yield", Statement("yield", lineno)
 
-        async def foo_nested_error():  # error: 8, "exit", Statement("yield", lineno+1)# error: 8, "exit", Statement("yield", lineno+1)
-            yield  # error: 12, "yield", Statement("function definition", lineno-1)# error: 12, "yield", Statement("function definition", lineno-1)
+        async def foo_nested_error():  # error: 8, "exit", Statement("yield", lineno+1)
+            yield  # error: 12, "yield", Statement("function definition", lineno-1)
```
jakkdl, line 165: Think previous behaviour here was incorrect, and these lines should only give one error each.

tests/eval_files/trio911.py
```diff
@@ -592,8 +592,14 @@ async def foo_boolops_1():  # error: 0, "exit", Stmt("yield", line+1)
...
-async def foo_boolops_2():  # error: 0, "exit", Stmt("yield", line+1) # error: 0, "exit", Stmt("yield", line+1)
-    _ = await foo() and (yield) and await foo() and (yield)
+async def foo_boolops_2():  # error: 0, "exit", Stmt("yield", line+4) # error: 0, "exit", Stmt("yield", line+6)
+    # known false positive - but chained yields in bool should be rare
+    _ = (
+        await foo()
+        and (yield)
+        and await foo()
+        and (yield)  # error: 13, "yield", Stmt("yield", line-2, 13)
+    )
```
jakkdl, line 602: The behaviour difference in 91X boolop handling, i.e. multiple yields with await between them in the same boolop statement will now error ... which really seems like it does not matter at all. But cases that actually matter, i.e. stuff where you're not cramming a ton of stuff into a single line, should be good as before.

tests/eval_files/trio911.py
```diff
@@ -742,6 +748,10 @@ async def foo_loop_static():
         await foo()
     yield  # error: 4, "yield", Stmt("yield", line-4)
 
+    for _ in """""":
+        await foo()
+    yield  # error: 4, "yield", Stmt("yield", line-4)
+
```
jakkdl, line 754: cst's way of treating literal strings includes what the string quotation marks are, so added this test.

tests/test_flake8_trio.py
```diff
@@ -366,7 +366,8 @@ def assert_correct_lines_and_codes(errors: Iterable[Error], expected: Iterable[E
                 sep=" | ",
                 file=sys.stderr,
             )
-    assert not any_error
+            error_count += abs(error_dict[line][code] - expected_dict[line][code])
+    assert error_count == 0
```
jakkdl, line 370: added to be able to see if I was actually making progress when fixing a bug and the number of errors go from a double-digit number to another double-digit number. Also for the fun of it

tests/test_flake8_trio.py
```diff
@@ -479,11 +481,11 @@ async def foo():
             if val is not None:
                 function_str += f"  {arg}:\n    {val}\n"
 
-        tree = ast.parse(function_str)
+        module = cst.parse_module(function_str)
 
         # not a type error per se, but it's pyright warning about assigning to a
         # protected class member - hence we silence it with a `type: ignore`.
-        plugin._tree = tree  # type: ignore
+        plugin._module = module  # type: ignore
```
jakkdl, line 488: 91X now uses libccst!

jakkdl (commented):
comments on the diff to ease review

flake8_trio/visitors/visitor91x.py
Zac-HD: I think you've got a comment in the code about this further down, but I think a similar short explanation here (1-2 lines?) would be nice.

Zac-HD (approved):
With tests passing, this is IMO a clear step forward from the status quo (🎉), so given the nit-picky nature of remaining comments I'll merge this now and we can follow up in a future PR - it'll be easier than re-reviewing this one.  Thanks again!

flake8_trio/visitors/helpers.py
jakkdl: you mean using `iter_guaranteed_once` ?
It has all the same tests that the ast version had, and vast majority of the logic is the same, so I'm pretty confident it's equivalent actually. And if/when 103/104 is converted to libcst there will be no ast users left using this.

flake8_trio/visitors/helpers.py
jakkdl: Oh, in the case of `iter_guaranteed_once` (literal parameter to range) it's actually plenty fine to use Integer/Float.evaluated_value; but I opened https://github.com/Instagram/LibCST/issues/878 for other cases.

tests/eval_files/trio910.py
jakkdl: Opened #144

flake8_trio/visitors/helpers.py
```diff
@@ -27,6 +28,7 @@
 
     T = TypeVar("T", bound=Flake8TrioVisitor)
     T_CST = TypeVar("T_CST", bound=Flake8TrioVisitor_cst)
+    T_EITHER = TypeVar("T_EITHER", Flake8TrioVisitor, Flake8TrioVisitor_cst)
```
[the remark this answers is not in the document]
jakkdl: Confused me for a minute why the multi-arg form doesn't give errors when passed subtypes, but figured out that it transforms the type.
And mypy gave me errors on the union one, but that's a bug with a PR in the pipe: https://github.com/python/mypy/issues/14755

flake8_trio/visitors/flake8triovisitor.py
jakkdl: opened https://github.com/Instagram/LibCST/issues/880

PR #145, opened 2023-02-28
code cleanup after 91x PR

jakkdl:
Fix comments from #143 and general minor code cleanup


jakkdl:
Actually, since I landed on `ast.literal_eval` being pretty much the same as matching against an integer + handling unary operation, I rewrote `iter_guaranteed_once_cst` not to need `cst_literal_eval`. It probably wouldn't be very hard at all to evaluate unary/binary ops either, and `range(10*23)` isn't *that* uncommon. Can open an issue for it

And with the only remaining user of it being visit_While_test ... and I don't think we care about handling `while [1, 2]: ...` (and can get it for free if libcst implements evaluated_value for containers), I rewrote that one as well .. and completely removed `cst_literal_eval`.

flake8_trio/visitors/helpers.py
```diff
@@ -28,7 +28,7 @@
 
     T = TypeVar("T", bound=Flake8TrioVisitor)
     T_CST = TypeVar("T_CST", bound=Flake8TrioVisitor_cst)
-    T_EITHER = TypeVar("T_EITHER", Flake8TrioVisitor, Flake8TrioVisitor_cst)
+    T_EITHER = TypeVar("T_EITHER", bound=Flake8TrioVisitor | Flake8TrioVisitor_cst)
```
Zac-HD, line 31: I think this `|` syntax requires Python 3.10?

flake8_trio/visitors/helpers.py
```diff
@@ -181,20 +167,23 @@ def iter_guaranteed_once_cst(iterable: cst.BaseExpression) -> bool:
...
-        except Exception:  # noqa: PIE786
+        values: list[int] = []
+        for arg_arg in ensure_type(iterable, cst.Call).args:
+            arg = arg_arg.value
+            if isinstance(arg, cst.UnaryOperation):
+                if not isinstance(arg.expression, cst.Integer):
+                    return False
+                value = arg.expression.evaluated_value
+                if isinstance(arg.operator, cst.Minus):
+                    value = -value
```
Zac-HD, line 178: Missing the `~` operator!  https://docs.python.org/3/reference/expressions.html#unary-arithmetic-and-bitwise-operations

flake8_trio/visitors/helpers.py
```diff
@@ -181,20 +167,23 @@ def iter_guaranteed_once_cst(iterable: cst.BaseExpression) -> bool:
...
+            elif isinstance(arg, cst.Integer):
+                value = arg.evaluated_value
+            else:
+                return False
+            values.append(value)
+        if len(values) not in (1, 2, 3):
             return False
+        return len(range(*values)) > 0
```
Zac-HD, line 186: Raises an error a lint time if it would at runtime!  In which case I think we want to return `False`, in case the exception is caught without a checkpoint occuring.

flake8_trio/visitors/visitor91x.py
```diff
@@ -405,8 +399,10 @@ def visit_While(self, node: cst.While | cst.For):
     # state after the loop same as above, and in addition the state at any break
     def visit_While_test(self, node: cst.While):
         # save state in case of nested loops
-        # ugly workaround since cst doesn't have literal_eval
-        if cst_literal_eval(node.test):
+        # One could plausibly just check for True here
+        if (isinstance(node.test, cst.Name) and node.test.value == "True") or (
```
Zac-HD, line 403:
```suggestion
        if m.matches(node.test, m.Name("True")) or (
```

tests/eval_files/trio910.py
```diff
@@ -537,4 +544,9 @@ async def foo_comprehension_1():
 
 # should error
 async def foo_comprehension_2():
-    [await foo() for x in range(10) if foo()]
+    [await foo() for x in range(10) if bar()]
+
+
+# should not error
+async def foo_comprehension_3():  # error: 0, "exit", Statement("function definition", lineno)
+    [... async for x in bar()]
```
Zac-HD, line 552:
```suggestion
# should not error, see https://github.com/Zac-HD/flake8-trio/issues/144
async def foo_comprehension_3():  # error: 0, "exit", Statement("function definition", lineno)
    [... async for x in bar()]
```

Zac-HD (approved):
Minor nitpicks but great overall, go ahead and merge once you've addressed to your satisfaction 😁

flake8_trio/visitors/helpers.py
jakkdl: I think it's fine due to `from __future__ import annotations`?
But regardless, tests run with <3.10, and I don't think we care about whether `pyright` running on 3.9 passes.

flake8_trio/visitors/helpers.py
jakkdl: Good catch! Although I suspect it's quite rare to invert bits in range loops :sweat_smile:

flake8_trio/visitors/helpers.py
jakkdl: Bringing it out to #147

jakkdl:
Reading through https://docs.python.org/3/library/stdtypes.html#range it mentioned `len` not working with `sys.maxsize` - so code now handles *very* long loops now as well :smiley: 
```python
    # length > sys.maxsize
    for i in range(27670116110564327421):
        await foo()
    yield
```

flake8_trio/visitors/helpers.py
Zac-HD: 🤦 it's fine at runtime because we're in an `if TYPE_CHECKING:` block and so it never actually runs, nevermind.

Zac-HD:
> Reading through https://docs.python.org/3/library/stdtypes.html#range it mentioned `len` not working with `sys.maxsize`

Heh, that note is my fault 😂  (Hypothesis is a great way to find edge cases...)

flake8_trio/visitors/helpers.py
jakkdl: haha I didn't realize that either, https://github.com/snok/flake8-type-checking doing it's job!

flake8_trio/visitors/helpers.py
jakkdl: And of course this checker is gonna reach that scale :grin:

jakkdl:
ooh, nice!

Issue #124, opened 2023-02-04
Switch backend to libCST (?)


jakkdl:
Moving from #70 

## what does the UX look like if we're not leaning on flake8 for that? I presume similar, so how much work is that to add?
Quickly looked at whether the flake8 interface could still work - and it's possible to still use flake8 without much problem. [You can indicate that you want the lines of a file](https://flake8.pycqa.org/en/latest/plugin-development/plugin-parameters.html#indicating-desired-data) and then that can be dumped into `libcst.parse_module()`. Flake8 does require plugins to accept one of `tree`, `logical_line` or `physical_line` as parameter in `__init__` for it to classify the plugin and run it at the correct time, so we'd have to still accept the `tree` parameter - but we can just throw it out and construct our own tree from the `lines` sent at the same time.

Somewhat hackish, but means that the consideration of ditching flake8 is pretty much fully independent.

This even means that the transition can even be gradual - where in the intermediate phase you have both an AST and a libCST tree, and run separate runners for them.

## concrete breakdown of what can be autofixed vs continue to emit lint warnings
TODO

## suggested breakdown of tasks into reviewable-sized PRs (e.g. "convert UX, switch over logic, add autofixes")

1. With a gradual transition, this becomes pretty simple
1. Convert one simple visitor to libCST, getting the basic infrastructure set up.
1. Convert a couple more visitors to libCST, to see that it works well.
1. Add autofixes to visitors, figuring out how the structure of that is supposed to work.
1. Finish up all the last visitors

jakkdl:
Leaving flake8 as a backend does not really seem worth it at the moment, it's added some headaches and complications at times - but we've managed to work around pretty much all of them so I don't currently see any particular upside in ditching it - unless new versions come out that break things (which isn't that unlikely tbh).
Only thing would be like usability / speed concerns for people that don't want to run it through flake8, or ability to specify version properly in pre-commit.

Zac-HD:
OK, I like the idea of an incremental migration pathway so let's stick with `flake8` for now - we can always revisit *after* all the substantive logic is converted over.

For autofixes, it'd be nice to hook that from `shed` in [the same way that it already hooks Hypothesis' codemods](https://github.com/Zac-HD/shed/blob/master/src/shed/_codemods.py).  I'm OK with saying "lint xor codemod" for any particular check, which is all I'd use, though ideally we could "lint all and codemod some" - not sure how that would be implemented though.

jakkdl:
[...]

PR #146, opened 2023-03-02
Now able to run as a standalone module or shell program, with limited functionality

jakkdl:
First steps towards having it run as a standalone program, which is required for autofixing.

* takes basic structure from shed, copying `_get_git_repo_root`.
* Add some magic to get argparsing to work both as standalone and as a plugin
* add `flake8_trio/__main__` to run as a module, and a `console_scripts` entry point
* moved out config testing from `test_flake8_trio`, and added a couple tests to test the above
  * crossing my fingers that git tests will magically work in the CI environment as well

Also:
* Remove `flake8-bandit` - running `git` directly is .. kinda sus https://flake8.codes/S607/ ; but if `shed` does it I'm not gonna bother, and I'm very much just disabling ~all errors bandit throws, so it's not really contributing anything.
* `reportUnusedCallResult` is frustrating, byebye~

Remaining steps:
* Color support - I was most of the way to replicating flake8's color printing, but once it got into testing/coverage and windows support I abandoned it and threw it out. For a later PR, if ever.
* Disable visitors with other ways than `--enable-visitor-codes-regex`
* pre-commit support
* read config
added these to OP in #124

### DIFF
--- a/flake8_trio/__init__.py
+++ b/flake8_trio/__init__.py
@@ -12,11 +12,14 @@ Pairs well with flake8-async and flake8-bugbear.
 from __future__ import annotations
 
 import ast
+import functools
 import keyword
 import os
 import re
+import subprocess
+import sys
 import tokenize
-from argparse import ArgumentTypeError, Namespace
+from argparse import ArgumentParser, ArgumentTypeError, Namespace
 from typing import TYPE_CHECKING
 
 import libcst as cst
@@ -37,6 +40,24 @@ if TYPE_CHECKING:
 __version__ = "23.2.5"
 
 
+# taken from https://github.com/Zac-HD/shed
+@functools.lru_cache
+def _get_git_repo_root(cwd: str | None = None) -> str:
+    return subprocess.run(
+        ["git", "rev-parse", "--show-toplevel"],
+        check=True,
+        timeout=10,
+        capture_output=True,
+        text=True,
+        cwd=cwd,
+    ).stdout.strip()
+
+
+@functools.cache
+def _should_format(fname: str) -> bool:
+    return fname.endswith((".py",))
+
+
 # Enable support in libcst for new grammar
 # See e.g. https://github.com/Instagram/LibCST/issues/862
 # wrapping the call and restoring old values in case there's other libcst parsers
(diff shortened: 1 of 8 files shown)


flake8_trio/__init__.py
```diff
@@ -53,6 +74,44 @@ def cst_parse_module_native(source: str) -> cst.Module:
...
+        metavar="file",
+        dest="files",
+        help="Files(s) to format, instead of autodetection.",
+    )
+    Plugin.add_options(parser)
+    args = parser.parse_args()
+    Plugin.parse_options(args)
+    if args.files:
+        # TODO: go through subdirectories if directory/ies specified
+        all_filenames = args.files
```
Zac-HD, line 90: 🤷‍♂️, I'm pretty happy to say that you have to pass exact files if you're passing files.

flake8_trio/__init__.py
```diff
@@ -53,6 +74,44 @@ def cst_parse_module_native(source: str) -> cst.Module:
...
+        except (subprocess.SubprocessError, FileNotFoundError):
+            print("Doesn't seem to be a git repo; pass filenames to format.")
+            sys.exit(1)
+        all_filenames = [
+            os.path.join(root, f) for f in all_filenames if _should_format(f)
+        ]
+    for file in all_filenames:
+        plugin = Plugin.from_filename(file)
+        for error in sorted(plugin.run()):
+            print(f"{file}:{error}")
```
Zac-HD, line 112: A small `ProcessPoolExecutor` invocation here can make it n_cores times faster for large repos... although we'd still want to sort by file and lineno.  Let's do that later.

flake8_trio/__init__.py
```diff
@@ -86,16 +145,23 @@ def run(self) -> Iterable[Error]:
...
     @staticmethod
-    def add_options(option_manager: OptionManager):
-        # Disable TRIO9xx calls
-        option_manager.extend_default_ignore(default_disabled_error_codes)
+    def add_options(option_manager: OptionManager | ArgumentParser):
+        if isinstance(option_manager, ArgumentParser):
+            # if run as standalone
+            add_argument = option_manager.add_argument
+        else:  # if run as a flake8 plugin
+            # Disable TRIO9xx calls
```
Zac-HD, line 153:
```suggestion
            # Disable TRIO9xx calls by default
```

Zac-HD (approved):
Merging this as the only comments here are tiny nitpicks and I'm really excited to see where it's going 🎉

flake8_trio/__init__.py
jakkdl: I figured, since that's how shed does it, but I would like to support as much of flake8's functionality as possible to ease any transition from running [name of this project] as a flake8 plugin to running it standalone. Ofc flake8 is kinda crazy so I won't bother supporting everything (e.g. `--extend-ignore`, `--ignore`, `--select`, `--extend-select` I think kinda just makes everything more complicated), but I think supporting directories shouldn't be very complicated and doesn't have much downsides.

PR #148, opened 2023-03-06
Capitalize errors

jakkdl:
noticed it randomly

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,23 +20,23 @@ pip install flake8-trio
 
 ## List of warnings
 
-- **TRIO100**: a `with trio.fail_after(...):` or `with trio.move_on_after(...):`
+- **TRIO100**: A `with trio.fail_after(...):` or `with trio.move_on_after(...):`
   context does not contain any `await` statements.  This makes it pointless, as
   the timeout can only be triggered by a checkpoint.
 - **TRIO101**: `yield` inside a nursery or cancel scope is only safe when implementing a context manager - otherwise, it breaks exception handling.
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -68,6 +68,15 @@ repos:
...
           - flake8-type-checking
 
+-   repo: https://github.com/PyCQA/flake8
+    rev: 6.0.0
+    hooks:
+    -   id: flake8
+        args: ["--exclude", ".*,tests/eval_files/*"]
+        language_version: python3
+        additional_dependencies:
+          - flake8-pyi
```
Zac-HD, line 78: This looks identical to the above, can we just add to the `additional_dependencies` list?

Zac-HD (approved):

.pre-commit-config.yaml
jakkdl: Oh oops, I didn't notice that got added to this PR. I did a cleanup of stashes and branches yesterday, and must've mixed up the playing-around-with-flake8-pyi stash/branch with the capitalize-errors stash/branch. I'll remove it from this PR and recheck what the end result of playing around with flake8-pyi was, there was some reason I didn't go ahead and push for its inclusion right away.

jakkdl: Oh, and re it being in a separate environment:
> Note: Be careful when using this plugin in the same environment as other flake8 plugins, as they might generate errors that are inappropriate for .pyi files (e.g., about missing docstrings). We recommend running flake8-pyi in a dedicated environment in your CI.

https://github.com/PyCQA/flake8-pyi
(which clearly should be labeled with a comment in the config)

PR #149, opened 2023-03-06
implement autofix for TRIO100

jakkdl:
Before jumping on autofixing TRIO91X I decided to start with the simpler TRIO100.
Turns out I'd forgotten how messy it is to manipulate code and preserve whitespace, comments, lines, etc etc etc :upside_down_face: 

~~But I remember solving it in shed, so I can probably crib that from there and write some helper function.
Also remaining is actually changing the file on disk, and adding a flag.~~

~~shed/Black/isort is also back to being incredibly annoying, I'll have to figure out a better way of writing autofix files so they don't get fucked with as much as is done currently.~~

### DIFF
--- a/flake8_trio/runner.py
+++ b/flake8_trio/runner.py
@@ -100,20 +100,20 @@ class Flake8TrioRunner(ast.NodeVisitor):
 
 
 class Flake8TrioRunner_cst:
-    def __init__(self, options: Namespace):
+    def __init__(self, options: Namespace, module: Module):
         super().__init__()
         self.state = SharedState(options)
         self.options = options
         self.visitors: tuple[Flake8TrioVisitor_cst, ...] = tuple(
             v(self.state) for v in ERROR_CLASSES_CST if self.selected(v.error_codes)
         )
+        self.module = module
 
-    def run(self, module: Module) -> Iterable[Error]:
+    def run(self) -> Iterable[Error]:
         if not self.visitors:
             return
-        wrapper = cst.MetadataWrapper(module)
         for v in self.visitors:
-            _ = wrapper.visit(v)
+            self.module = cst.MetadataWrapper(self.module).visit(v)
         yield from self.state.problems
 
     def selected(self, error_codes: dict[str, str]) -> bool:
--- a/flake8_trio/visitors/helpers.py
+++ b/flake8_trio/visitors/helpers.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 
 import ast
 from fnmatch import fnmatch
-from typing import TYPE_CHECKING, NamedTuple, TypeVar
+from typing import TYPE_CHECKING, NamedTuple, TypeVar, cast
 
 import libcst as cst
 import libcst.matchers as m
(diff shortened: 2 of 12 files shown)


jakkdl:
It be done!

Very happy with this new autofix testing infrastructure, and I even just now realized that I can exclude autofix files from being shed'd by excluding them in pre-commit ^^

Uses a similar approach to https://github.com/Zac-HD/shed/pull/52 ; and shed would normally codemod away the useless pass statements that are inserted in the autofix files sometimes ... which would cause the autofix tests to fail.
It would be great if I could highly selectively disable `pass` removal, so shed can check that the autofixed files are otherwise generated with good formatting. `fmt: off` does nothing.

`--generate-autofix` maybe deserves a mention in `CONTRIBUTING.md` - it helps a lot when developing. And ofcourse `--autofix` should be documented, but it should probably be a comma-separated list anyway.

flake8_trio/runner.py
```diff
@@ -100,20 +100,20 @@ def visit(self, node: ast.AST):
 
 
 class Flake8TrioRunner_cst:
```
jakkdl, line 102: the runner is kind of .. not necessary and clunky, esp as currently `__init__` does a bunch of logic itself. Might warrant some restructuring.

flake8_trio/runner.py
```diff
@@ -100,20 +100,20 @@ def visit(self, node: ast.AST):
...
 
-    def run(self, module: Module) -> Iterable[Error]:
+    def run(self) -> Iterable[Error]:
         if not self.visitors:
             return
-        wrapper = cst.MetadataWrapper(module)
         for v in self.visitors:
-            _ = wrapper.visit(v)
+            self.module = cst.MetadataWrapper(self.module).visit(v)
         yield from self.state.problems
```
jakkdl, line 117: now saves and updates self.module after visiting.

tests/eval_files/trio100.py
```diff
@@ -3,20 +3,23 @@
 import trio
 
 with trio.move_on_after(10):  # error: 5, "trio", "move_on_after"
-    pass
+    ...
```
jakkdl, line 6: replaced `pass` with ellipsis partly in attempts to avoid shed headaches, but means you can see a difference with ellipsis's transferred over, vs injected `pass`s, in the autofix file.

tests/autofix_files/trio100.py
```diff
@@ -0,0 +1,84 @@
+# type: ignore
```
jakkdl, line 1: there's not a ton to see here ("trio100_simple_autofix.py" is easier to read), but if you want to review it you should view this and `tests/eval_files/trio100.py` with a difftool of your choice.

tests/eval_files/trio100_simple_autofix.py
```diff
@@ -0,0 +1,51 @@
...
+    # b
+    # c
+    trio.move_on_after(  # error: 4, "trio", "move_on_after"
+        # e
+        9999999999999999999999999999999999999999999999999999999  # f
+        # g
+    )  # h
+    # i
+):  # this comment is kept
+    ...
```
jakkdl, line 51: I can maybe detect this and abort the autofix, rather than eating the comments. But I'm not sure if this happens ever? Maybe if you're passing a lot of parameters to one of fail_after/fail_at/move_on_after/CancelScope so it's gotten split across lines by Black.

tests/test_config_and_args.py
```diff
@@ -63,8 +63,24 @@ def test_run_no_git_repo(
...
+
+
+def test_run_100_autofix(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch, capsys: pytest.CaptureFixture[str]
+):
+    err_msg = _common_error_setup(tmp_path)
+    monkeypatch.chdir(tmp_path)
+    monkeypatch.setattr(
+        sys, "argv", [tmp_path / "flake8_trio", "--autofix", "./example.py"]
+    )
```
jakkdl, line 77: might deserve a helper function, and/or helper fixture

tests/test_flake8_trio.py
```diff
@@ -65,19 +69,40 @@ def check_version(test: str):
...
         expected = []
 
     plugin = Plugin.from_source(content)
     _ = assert_expected_errors(plugin, *expected, args=parsed_args)
 
+    check_autofix(test, plugin, generate_autofix)
+
+
+def test_assure_all_autofix_files_used():
+    assert set(autofix_files.keys()) - checked_autofix == set()
```
jakkdl, line 104: this one is a minor pain since I usually develop locally with `--failed-first`, but didn't see any simple way of forcing this to be run late without installing pytest plugins (and that might still get overridden by `--failed-first`). But should help the occasional messed up file management by making sure all files in autofixed_files are used.
(although ig that could be done by static analysis of directory contents....)

jakkdl (commented):
Some comments to ease reviewing

tests/eval_files/trio100_simple_autofix.py
jakkdl: Fixed, comments are now preserved.

tests/test_flake8_trio.py
jakkdl: fixed, now statically checking the contents of the file lists.

tox.ini
```diff
@@ -37,9 +37,9 @@ source =
 
 [flake8]
 max-line-length = 90
-extend-ignore = S101, D101, D102, D103, D105, D106, D107, TC006
+extend-ignore = S101, D101, D102, D103, D105, D106, D107
```
jakkdl, line 40: finally realized why they want to stringify types in `cast` - the types aren't going to be imported at runtime. And is probably a minor speedup as well

flake8_trio/visitors/helpers.py
```diff
@@ -341,3 +341,62 @@ def func_has_decorator(func: cst.FunctionDef, *names: str) -> bool:
             ),
         )
     )
+
+
+def get_comments(node: cst.CSTNode | Iterable[cst.CSTNode]) -> Iterator[cst.EmptyLine]:
+    # pyright can't use hasattr to narrow the type, so need a bunch of casts
```
jakkdl, line 347: mypy manages it without problem though...

jakkdl:
I realize now I added this comment in #124 
> just remove the context, unless there's other warnings about the content of the context.

that's ... really really hard to implement properly - at least until everything is converted to libcst. So that will have to wait for a later PR, if it turns out to actually be of use. I was probably mainly thinking of missing `await`s - but if I can fix trio typing so type checkers can warn on async function calls not being awaited that should be a non-issue.

flake8_trio/visitors/visitor100.py
```diff
@@ -46,12 +51,16 @@ def visit_With(self, node: cst.With) -> None:
...
+        self, original_node: cst.With, updated_node: cst.With
+    ) -> cst.BaseStatement | cst.FlattenSentinel[cst.BaseStatement]:
         if not self.has_checkpoint_stack.pop():
             for res in self.node_dict[original_node]:
                 self.error(res.node, res.base, res.function)
-        # if: autofixing is enabled for this code
-        # then: remove the with and pop out it's body
+
+            if self.options.autofix and len(updated_node.items) == 1:
+                return flatten_preserving_comments(updated_node)
```
Zac-HD, line 62: I assume that we don't actually emit the error for autofixed code?

Zac-HD (approved):
This all looks good to me!  Rebase on master and then merge?

flake8_trio/visitors/visitor100.py
jakkdl: This currently always does, I'll introduce ways of configuring that soon :)

jakkdl:
merged!

PR #150, opened 2023-03-07
Add flake8-pyi to pre-commit

jakkdl:
So this doesn't currently complain about anything in the `.pyi` files - I don't remember whether that is because I addressed all of that in a previous commit (`git blame typings/flake8/options/manager.pyi` shows that file has been modified three times (none of the other has)) or when generating the files, or if this never complained about anything with the generated files.

But regardless I don't think the plugin does much, if anything, for us, we only use stubs against flake8 (which is partly on it's way out), and the stubs aren't exposed to the end user in any way. But since I accidentally pushed it in #148 I thought I might as well bring it up here


jakkdl:
Wait no, there's some issue where flake8-pyi isn't run in pre-commit for some reason. Running it standalone gives a bunch of errors

Zac-HD (approved):
Makes sense and the comments make it clear!

Merge if and when you think it's ready 😊

.pre-commit-config.yaml
```diff
@@ -48,7 +48,7 @@ repos:
     rev: 6.0.0
     hooks:
     -   id: flake8
-        args: ["--exclude", ".*,tests/eval_files/*"]
+        types: ["python", "pyi"]
```
jakkdl, line 51: I .. got incredibly confused and thought flake8 was on it's way to including `.pyi` by default (see https://github.com/PyCQA/flake8/issues/1806 :no_mouth:) and went ahead and included it; but even after realizing it's not meant to handle `.pyi` by default I realized there weren't that many errors in there anyway - so not much reason to have a separate environment, can just handle that with per-file-ignores.

tox.ini
```diff
@@ -38,7 +38,16 @@ source =
 [flake8]
 max-line-length = 90
 extend-ignore = S101, D101, D102, D103, D105, D106, D107, TC006
-per-file-ignores = flake8_trio/visitors/__init__.py: F401, E402
+extend-enable = TC10
```
jakkdl, line 41: https://github.com/snok/flake8-type-checking#choosing-how-to-handle-forward-references

jakkdl (commented):
went down a rabbit hole with figuring out this, so changes are worthy of some comments.

Zac-HD (approved):
Awesome work!

PR #153, opened 2023-03-10
split shed, format&sort toml+yaml

jakkdl:
Split out shed for more granular control of the underlying tools, also added auto-formatting for yaml&toml and other minor config changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,58 +1,59 @@
+---
 name: flake8-trio CI
 
 on:
-  push:
-    branches: [ main ]
-  pull_request:
-    branches: [ main ]
+    push:
+        branches: [main]
+    pull_request:
+        branches: [main]
 
 jobs:
-  check:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v3
-    - name: Set up Python 3.11
-      uses: actions/setup-python@v3
-      with:
-        python-version: "3.11"
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip setuptools pre-commit
-    - name: Run checks
-      run: |
-        pre-commit run --all-files
+    check:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v3
+            - name: Set up Python 3.11
+              uses: actions/setup-python@v3
+              with:
+                  python-version: '3.11'
+            - name: Install dependencies
+              run: |
+                  python -m pip install --upgrade pip setuptools pre-commit
+            - name: Run checks
+              run: |
+                  pre-commit run --all-files
 
-  test:
-    runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        python-version: ["3.9", "3.10", "3.11"]
-      fail-fast: false
-    steps:
-    - uses: actions/checkout@v3
-    - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v3
-      with:
-        python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
-      run: |
-        python -m pip install --upgrade pip setuptools tox
-        python -m tox --notest --recreate -e flake8_5,flake8_6
-    - name: Run tests
-      run: python -m tox -e flake8_5,flake8_6
+    test:
+        runs-on: ubuntu-latest
+        strategy:
+            matrix:
+                python-version: ['3.9', '3.10', '3.11']
+            fail-fast: false
+        steps:
+            - uses: actions/checkout@v3
+            - name: Set up Python ${{ matrix.python-version }}
+              uses: actions/setup-python@v3
+              with:
+                  python-version: ${{ matrix.python-version }}
+            - name: Install dependencies
+              run: |
+                  python -m pip install --upgrade pip setuptools tox
+                  python -m tox --notest --recreate -e flake8_5,flake8_6
+            - name: Run tests
+              run: python -m tox -e flake8_5,flake8_6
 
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -1,58 +1,59 @@
+---
```
jakkdl, line 1: no changes, just reformatted.

.pre-commit-config.yaml
```diff
@@ -1,81 +1,110 @@
...
-          - pytest
-          - flake8
-          - trio
-          - anyio
+                - anyio
+                - flake8
+                - hypothesis
+                - hypothesmith
+                - pytest
+                - trio
```
jakkdl, line 44: sorted

.pre-commit-config.yaml
```diff
@@ -1,81 +1,110 @@
...
-    -   id: fix-encoding-pragma
-        args: [--remove]
-    -   id: check-yaml
-    -   id: debug-statements
-        language_version: python3
+    - repo: https://github.com/pre-commit/pre-commit-hooks
+      rev: v4.4.0
+      hooks:
+          - id: check-merge-conflict
+          - id: check-toml
```
jakkdl, line 57: sorted, and added these two

.pre-commit-config.yaml
```diff
@@ -1,81 +1,110 @@
...
-          - flake8-pytest-style
-          - flake8-return
-          - flake8-simplify
-          - flake8-type-checking
-          - flake8-pyi
+    - repo: https://github.com/PyCQA/flake8
+      rev: 6.0.0
+      hooks:
+          - id: flake8
+            types_or: [python, pyi]
```
jakkdl, line 71: *very* important that this is `types_or` and not `types`, else it only checked stub files :sweat_smile:

.pre-commit-config.yaml
```diff
@@ -1,81 +1,110 @@
...
+                - flake8-datetimez
+                - flake8-docstrings
+                - flake8-mutable
+                - flake8-noqa
+                - flake8-pie
+                - flake8-pyi
+                - flake8-pytest-style
+                - flake8-return
+                - flake8-simplify
+                - flake8-type-checking
```
jakkdl, line 87: only sorted, turns out there were duplicates!

.pre-commit-config.yaml
```diff
@@ -1,81 +1,110 @@
...
+    - repo: https://github.com/jumanjihouse/pre-commit-hook-yamlfmt
+      rev: 0.2.2
+      hooks:
+          - id: yamlfmt
+
+    - repo: https://github.com/macisamuele/language-formatters-pre-commit-hooks
+      rev: v2.7.0
+      hooks:
+          - id: pretty-format-toml
+            args: [--autofix]
```
jakkdl, line 110: added

pyproject.toml
```diff
@@ -1,13 +1,30 @@
-[tool.pyright]
```
jakkdl, line 1: file sorted, and autoflake+isort added, pyright&codespell unchanged

tests/test_all_visitors_imported.py
```diff
@@ -16,8 +16,17 @@ def test_all_visitors_imported():
...
+            if m := re.fullmatch(r"from \. import \(\n", line):
+                in_import = True
+            elif in_import and (m := re.fullmatch(r" *(?P<module>\w*),\n", line)):
                 visited_files.add(m.group("module"))
+            elif in_import and re.fullmatch(r"\)\n", line):
+                in_import = False
+
+    # check that parsing succeeded
+    assert in_import is False
+
```
jakkdl, line 31: required new logic with formatting change in the `__init__.py`

tox.ini
```diff
@@ -35,18 +35,6 @@ source =
     flake8_trio
     */site-packages/flake8_trio
 
-[flake8]
```
jakkdl, line 38: moved so the coverage sections were next to eachother, and removed comment about Y021/Y048

Zac-HD:
What's the problem with just using `shed`?  Probably obvious, but I really like having a single standard configuration across all my projects.

Could the issue be addressed by disabling `shed` for some test files and _only_ using `black` there?

jakkdl:
The current issue is very specifically that I want to disable `autoflake` removing `pass` for `autofix_files`. The particular checks I *want* to run for those files I don't fully know, and while just running Black would probably be fine, I would prefer to default to running everything on the semi-generated code to alert me of any possible weird stuff going on asap, so I can fix it or be aware of the limitation. I could maybe also/instead stop blanket disabling flake8 for eval_ & autofix_files.

But given how much I've fought with shed in the past with `eval_files`, and even occasionally with e.g. the imports in `visitors/__init__.py`, I personally prefer to have a bit more control than shed currently allows. (but definitely not as much as e.g. isort, that's just a crazy amount of config :sweat_smile:) So yeah, that's why I've asked for any ways of making marginal changes to shed's behaviour.

It's also sometimes been frustrating to track down *why* a diff is generated, when shed modifies a file I either have to debug, or manually run the underlying tools, if/when I want to know what tool it is that generated the diff. Sometimes this is for requesting bugs/features in those tools/shed, or when I want to know whether to add `# noqa`, `# isort: skip` and/or `fmt: off`.

tl;dr, we could def go back to shed if you insist - but I do find it frustrating/limiting at times, aside from the current autoflake `pass` problem which prompted this.

Zac-HD (approved):
Makes sense!

I'm always a bit 😥 when `shed` _isn't_ the right choice, but in this case I agree it's better to use the underlying tools directly.

PR #154, opened 2023-03-10
add comprehension support to 91x

### LINKED ISSUE #144: TRIO91X: Support comprehensions
Comprehensions are currently completely invisible to 91X, so these are currently misbehaving:
```python
async def foo_comprehension_1():
    [... for x in range(10) if await foo()]


# should error
async def foo_comprehension_2():
    [await foo() for x in range(10) if bar()]

# should not error
async def foo_comprehension_3():  # error: 0, "exit", Statement("function definition", lineno)
    [... async for x in bar()]
```

One cannot yield or return inside a comprehension though, so will require separate logic from loops - but it should be much simpler.
I'm pretty sure an async generator or `for` *cannot* be guaranteed to iterate at least once either, even with liberal use of starred expressions and the like.

https://peps.python.org/pep-0530/
https://libcst.readthedocs.io/en/latest/nodes.html#comprehensions

jakkdl:
fixes #144 

Fairly straightforward implementation, other than the gnarly logic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 *[CalVer, YY.month.patch](https://calver.org/)*
 
+## Future
+- TRIO91X now supports comprehensions
+
 ## 23.2.5
 - Fix false alarms for `@pytest.fixture`-decorated functions in TRIO101, TRIO910 and TRIO911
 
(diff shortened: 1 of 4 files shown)


Zac-HD (approved):
👍 fiddly but very nicely explained and tested 🚢

Issue #155, opened 2023-03-13
reportUnusedCoroutine / unused-coroutine *is* in fact raised for trio


jakkdl:
if trio_typing is installed, *and the call is reachable*. I had no clue that both mypy and pyright will straight up not check unreachable statements, but after moving out `sleep_forever` and the `serve` functions in https://github.com/Zac-HD/flake8-trio/blob/main/tests/eval_files/trio105.py to separate functions:
```sh
$ mypy tests/eval_files/trio105.py | grep -c unused-coroutine
21
$ pyright tests/eval_files/trio105.py | grep -c reportUnusedCoroutine
4
```
the disparity is due to pyright not raising it if there's other problems with the call (missing parameters etc), when I fixed that and switch to a normal way of opening nurseries they both jump up to 26. The total number of TRIO105 errors defined in the file is 29, and the differences are all logical (1. `open_process` only exists on win32 and trio-typing currently doesn't export it in the stubs at all, so mypy/pyright says it doesn't exist, 2. `async with trio.open_file(...)` complains about missing `__aenter__` instead, and 3. `k = trio.open_file(par)` being an error is a limitation in the flake8-trio implementation).

tl;dr: turns out that I due to a combination of confusions was wrong! So if end users are running with type-checking and has stubs from trio_typing, TRIO105 should be fully redundant both for trio and anyio, and can be removed.

Without trio_typing ~no errors are raised since trio doesn't have `py.typed` marker, if it had I think it would work fully as well.

Zac-HD:
I’d actually like to keep what we already have for now; it’s unfortunately pretty common to disable typechecking so a simple lint isn’t quite redundant.

I'm also hoping that the static-typing-for-Trio situation will improve over the coming months, e.g. by merging trio-typing into Trio proper and adding tests that ensure it's all up to date (e.g. fixing [1] and [2] above).  Once/if that's in, we probably _can_ remove `TRIO105` 🙂

jakkdl:
No need to hope, it's just a matter of prioritizing trio static typing and I'll get it done :grin: 
Sounds good!

PR #157, opened 2023-03-14
add mypy to get warnings on unreachable code, fix/ignore various small typing errors

jakkdl:
Didn't see an option to warn on unreachable code in pyright, and I tried running Vulture for that but that started requiring me to disable 90%+ of it since it doesn't detect that visitor/leave functions are run, and a lot of other weirdness.
So I settled for running .. another type checker :partying_face: But disabled a couple error codes for now.
It did give some decent warnings on eval files though:
* 103: changing if statements' test from `True` to `...` makes it clearer that the test isn't evaluated
* 104: infinite loops at module level turned off type checking (see #155 )
* 105: noreturn calls have to be split out to get type checking

It does feel silly to run double type checkers, but I'm progressively realizing that linting the h*ck out of the eval files is actually useful - despite how irritating it can be.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,6 +28,11 @@ repos:
       hooks:
           - id: isort
 
+    - repo: https://github.com/pre-commit/mirrors-mypy
+      rev: v1.1.1
+      hooks:
+          - id: mypy
+
     - repo: https://github.com/RobertCraigie/pyright-python
       rev: v1.1.298
       hooks:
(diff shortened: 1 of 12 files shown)


Zac-HD (approved):

Zac-HD:
> I'm progressively realizing that linting the h*ck out of the eval files is actually useful - despite how irritating it can be.

Oooh, yes.  One of the few things more annoying than a really strict linter is realizing that you should have started using one a while ago!

jakkdl:
haha! so true, so true

Issue #70, opened 2022-11-26
Idea: warn on unnecessary checkpoints added to avoid TRIO107/108


Zac-HD:
I've noticed a few places where `await trio.sleep(0)` is added in places that don't actually need a checkpoint, and so I've thought up a two-part plan to mitigate cargo-cult copy-paste coding.

1. Instead of `await trio.sleep(0)`, suggest `await trio.lowlevel.checkpoint()`.  It's an identical runtime effect, but the name is *much* more suggestive of what's actually going on: "we need a checkpoint here for some low-level reason" ✅
2. In each async function, for each `await trio.lowlevel.checkpoint()`, check if there would be no 107/108 warnings emitted if that checkpoint was *removed* (in addition to any already-suggested-removals, of course).

This is marked as an "idea" issue rather than for implementation because while I'm fairly confident (1) is a good plan, (2) seems a little harder to implement.  It's also triggering my "this feels pretty tedious" detector, but building a libCST-based autofixer also seems like a little too much duplication of effort, right?  Even if we *could* autofix 105, 107, 108, 112, and 113; and hook it in to `shed --refactor`... tempting but *probably* not worth it 😕 

Thoughts?

jakkdl:
Yeah 1 sounds great.

2. I don't think would be that hard to implement, given how .. *extreme* .. the 107/108 implementation is already. `self.uncheckpointed_statements` would instead of a set be a list sets, and when the code previously emptied it, it can instead push a new empty set on the stack and do some magic logic.
It's possible the trickier cases with loops is harder and not worth it, if that's the case it seems totally fine not to warn about unnecessary checkpoints in that case.
I guess it's harder when you combine it with other warnings that tell you to refactor, if not impossible without libCST, but it doesn't seem unreasonable to expect users to ignore errors from that when combined with other ones, or could be more explicit and mention that in the error text or even suppress the error if there are other ones in the vicinity.

But biggest downside is probably not implementation time, but that the class is already 300 lines long and I do not envy anybody trying to understand it (imagine a bug turns up in `visit_loop` :scream:) and adding more functionality to it would certainly not make it easier to maintain.
[...]

PR #158, opened 2023-03-15
implement autofix for 91x

jakkdl:
does *still* not finish #70 :sweat_smile: It "merely" inserts checkpoints wherever needed. 

TODO:
- [x] insert checkpoints
- [x] add import if necessary
- [x] insert anyio checkpoint if anyio is the preferred library (and anyio import)
- [x] handle multiple yields on the same line (but they're not autofixed)
- [x] save state on new class members
- [x] add some more comments
- [x] add review comments

And added a couple more tasks to #124 before this is reasonably usable by an end user. That includes checks for whether autofix toggling actually is toggled, which I kinda know isn't done properly by this anyway (but atm it mostly doesn't matter cause the new file doesn't get written without the autofix flag). But I'll leave those for another PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -32,6 +32,9 @@ To check that all codes are tested and documented there's a test that error code
 ## Test generator
 Tests are automatically generated for files in the `tests/eval_files/` directory, with the code that it's testing interpreted from the file name. The file extension is split off, if there's a match for for `_py\d*` it strips that off and uses it to determine if there's a minimum python version for which the test should only run.
 
+### autofix files
+Checks that have autofixing can have a file in the `tests/autofix_files` directory matching the filename in `tests/eval_files`. The result of running the checker on the eval file with autofix enabled will then be compared to the content of the autofix file and will print a diff (if `-s` is on) and assert that the content is the same. `--generate-autofix` is added as a pytest flag to ease development, which will print a diff (with `-s`) and overwrite the content of the autofix file. Also see the magic line marker `pass # AUTOFIX_LINE ` below
+
 ### `error:`
 Lines containing `error:` are parsed as expecting an error of the code matching the file name, with everything on the line after the colon `eval`'d and passed as arguments to `flake8_trio.Error_codes[<error_code>].str_format`. The `globals` argument to `eval` contains a `lineno` variable assigned the current line number, and the `flake8_trio.Statement` namedtuple. The first element after `error:` *must* be an integer containing the column where the error on that line originates.
 #### `TRIOxxx:`
--- a/flake8_trio/runner.py
+++ b/flake8_trio/runner.py
@@ -13,7 +13,12 @@ from typing import TYPE_CHECKING
 
 import libcst as cst
 
-from .visitors import ERROR_CLASSES, ERROR_CLASSES_CST, utility_visitors
+from .visitors import (
+    ERROR_CLASSES,
+    ERROR_CLASSES_CST,
+    utility_visitors,
+    utility_visitors_cst,
+)
 
 if TYPE_CHECKING:
     from argparse import Namespace
@@ -104,6 +109,13 @@ class Flake8TrioRunner_cst:
         super().__init__()
         self.state = SharedState(options)
         self.options = options
+
+        # Could possibly enable/disable utility visitors here, if visitors declared
+        # dependencies
+        self.utility_visitors: tuple[Flake8TrioVisitor_cst, ...] = tuple(
+            v(self.state) for v in utility_visitors_cst
+        )
+
         self.visitors: tuple[Flake8TrioVisitor_cst, ...] = tuple(
             v(self.state) for v in ERROR_CLASSES_CST if self.selected(v.error_codes)
         )
@@ -112,7 +124,7 @@ class Flake8TrioRunner_cst:
     def run(self) -> Iterable[Error]:
         if not self.visitors:
             return
-        for v in self.visitors:
+        for v in (*self.utility_visitors, *self.visitors):
             self.module = cst.MetadataWrapper(self.module).visit(v)
         yield from self.state.problems
 
--- a/flake8_trio/visitors/visitor91x.py
+++ b/flake8_trio/visitors/visitor91x.py
@@ -8,7 +8,7 @@ each yield.
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 import libcst as cst
 import libcst.matchers as m
@@ -23,6 +23,10 @@ from .helpers import (
     func_has_decorator,
     iter_guaranteed_once_cst,
 )
+
+if TYPE_CHECKING:
+    from collections.abc import Sequence
+
 
 ARTIFICIAL_STATEMENT = Statement("artificial", -1)
 
@@ -46,8 +50,10 @@ class LoopState:
 
     uncheckpointed_before_continue: set[Statement] = field(default_factory=set)
     uncheckpointed_before_break: set[Statement] = field(default_factory=set)
-    artificial_errors: set[cst.Return | cst.FunctionDef | cst.Yield] = field(
-        default_factory=set
+
+    artificial_errors: set[cst.Return | cst.Yield] = field(default_factory=set)
+    nodes_needing_checkpoints: list[cst.Return | cst.Yield] = field(
+        default_factory=list
     )
 
     def copy(self):
@@ -58,6 +64,7 @@ class LoopState:
             uncheckpointed_before_continue=self.uncheckpointed_before_continue.copy(),
             uncheckpointed_before_break=self.uncheckpointed_before_break.copy(),
             artificial_errors=self.artificial_errors.copy(),
+            nodes_needing_checkpoints=self.nodes_needing_checkpoints.copy(),
         )
 
 
@@ -75,6 +82,12 @@ class TryState:
             except_uncheckpointed_statements=self.except_uncheckpointed_statements.copy(),
             added=self.added.copy(),
         )
+
+
+def checkpoint_statement(library: str) -> cst.SimpleStatementLine:
+    return cst.SimpleStatementLine(
+        [cst.Expr(cst.parse_expression(f"await {library}.lowlevel.checkpoint()"))]
+    )
 
 
 @error_class_cst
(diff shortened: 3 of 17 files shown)


Zac-HD:
Exciting!

flake8_trio/visitors/visitor91x.py
```diff
@@ -186,38 +204,103 @@ def visit_FunctionDef(self, node: cst.FunctionDef) -> None:
...
         # avoid duplicate error messages
         self.uncheckpointed_statements = set()
 
+        # return original node to avoid problems with identity equality
+        assert original_node.deep_equals(updated_node)
+        return original_node
+
+    # TODO: generate an error in these two if transforming+visiting is done in a single
+    # pass and emit-error-on-transform can be enabled/disabled. The error can't be
+    # generated in the yield/return since it doesn't know if it will be autofixed.
```
jakkdl, line 261: unlikely I'll go with that design, transform+visit in single pass I'm pretty sure would guarantee that errors are output with incorrect line markers so I think you have to first transform, then revisit without transform. But would be substantial runtime reduction if you can do in single pass ofc, so will have to play around a bit with it.

flake8_trio/visitors/visitor_utility.py
```diff
@@ -116,3 +119,22 @@ def visit_Import(self, node: ast.Import):
             name = alias.name
             if name in ("trio", "anyio") and alias.asname is None:
                 self.add_library(name)
+
+
+@utility_visitor_cst
```
jakkdl, line 124: pretty much just a copy of the ast version

tests/test_flake8_trio.py
jakkdl, line 1: remaining todo that [selectively] disabling autofix works, but now testing autofix for both trio and anyio.

flake8_trio/visitors/visitor91x.py
```diff
@@ -494,10 +589,31 @@ def leave_While_orelse(self, node: cst.While | cst.For):
 
         # reset break & continue in case of nested loops
         self.outer[node]["uncheckpointed_statements"] = self.uncheckpointed_statements
-        self.restore_state(node)
```
jakkdl, line 497: needed to be moved to `leave_While` since it has to be done after inserting checkpoints in the loop body.

flake8_trio/visitors/visitor91x.py
```diff
@@ -186,38 +207,112 @@ def visit_FunctionDef(self, node: cst.FunctionDef) -> None:
...
+
         self.restore_state(original_node)
-        return updated_node
+        return updated_node  # noqa: R504
+
+    # error if function exit/return/yields with uncheckpointed statements
+    # returns a bool indicating if any real (i.e. not artificial) errors were raised
+    # so caller can insert checkpoint before statement (if yield/return) or at end
+    # of body (functiondef)
+    def check_function_exit(
```
jakkdl, line 230: moved all artificial statement logic from error_91x to check_function_exit

flake8_trio/visitors/visitor91x.py
```diff
@@ -394,6 +493,8 @@ def visit_While(self, node: cst.While | cst.For):
         self.loop_state = LoopState()
         self.infinite_loop = self.body_guaranteed_once = False
 
+    visit_For = visit_While
```
jakkdl, line 496: ooooops, pretty nasty line to have forgotten about. Added a test that would've caught it.

flake8_trio/visitors/visitor91x.py
```diff
@@ -432,23 +533,32 @@ def visit_While_body(self, node: cst.While | cst.For):
...
-        )
-        for err_node in self.artificial_errors:
+        # uncheckpointed_before_continue, and checkpoints at the end of the loop
+        any_error = False
+        for err_node in self.loop_state.artificial_errors:
             for stmt in (
-                new_uncheckpointed_statements | self.uncheckpointed_before_continue
+                self.outer[node]["uncheckpointed_statements"]
+                | self.uncheckpointed_statements
+                | self.uncheckpointed_before_continue
```
jakkdl, line 548: got confused by past self and cleaned up logic a bit

flake8_trio/visitors/visitor91x.py
```diff
@@ -590,3 +721,77 @@ def visit_CompFor(self, node: cst.CompFor):
     # ignore their content
     def visit_GeneratorExp(self, node: cst.GeneratorExp):
         return False
+
+    def visit_Import(self, node: cst.Import):
+        """Register explicitly imported library.
+
+        This is *slightly* different enough from what the utility visitor does
+        that it's added here, but the functionality is maybe better placed in there.
```
jakkdl, line 729: it'd probably just need another bit saved in self.library on whether it's been explicitly imported, or just assumed/specified with `--anyio`. But I'll procrastinate on that with the on-the-spot-made-up-reasoning that there's ast visitors that use the current self.library functionality and I neither wanna update both nor have different library logic between them.

tests/eval_files/trio911.py
```diff
@@ -292,6 +303,22 @@ async def foo_while_endless_2():  # error: 0, "exit", Statement("function defini
...
+    await foo()
+
+
+async def foo_while_endless_4():
+    await foo()
+    while True:
+        yield
+        while True:
+            await foo()
+            yield
```
jakkdl, line 319: I don't remember why I added these :innocent:

tests/test_flake8_trio.py
```diff
@@ -341,9 +360,9 @@ def print_first_diff(errors: Sequence[Error], expected: Sequence[Error]):
     for err, exp in zip(errors, expected):
         if err == exp:
             continue
-        if not first_error_line or err.line == first_error_line[0]:
+        if not first_error_line or err.line == first_error_line[0].line:
             first_error_line.append(err)
-        if not first_expected_line or exp.line == first_expected_line[0]:
+        if not first_expected_line or exp.line == first_expected_line[0].line:
```
jakkdl, line 365: have never worked as intended since it was written I'm pretty sure :sweat_smile:

CONTRIBUTING.md
```diff
@@ -32,6 +32,9 @@ To check that all codes are tested and documented there's a test that error code
 ## Test generator
 Tests are automatically generated for files in the `tests/eval_files/` directory, with the code that it's testing interpreted from the file name. The file extension is split off, if there's a match for for `_py\d*` it strips that off and uses it to determine if there's a minimum python version for which the test should only run.
 
+### autofix files
+Checks that have autofixing can have a file in the `tests/autofix_files` directory matching the filename in `tests/eval_files`. The result of running the checker on the eval file with autofix enabled will then be compared to the content of the autofix file and will print a diff (if `-s` is on) and assert that the content is the same. `--generate-autofix` is added as a pytest flag to ease development, which will print a diff (with `-s`) and overwrite the content of the autofix file. Also see the magic line marker `pass # AUTOFIX_LINE ` below
```
Zac-HD, line 36: Instead of spitting out the resulting `.py` files, can we save the diff between before and after as a `.py.diff`?  It's much easier to get a sense of "what it's doing" in this format, and I'm concerned that without that code review is not going to be effective.

(although instead of constructing a three-way diff, I'd just let Pytest's assertion helper show the diff between the actual and expected diffs)

((diff diff diff, it hardly seems to ~~diff~~ mean anything anymore.  diff)

flake8_trio/runner.py
```diff
@@ -112,7 +125,7 @@ def __init__(self, options: Namespace, module: Module):
     def run(self) -> Iterable[Error]:
         if not self.visitors:
             return
-        for v in self.visitors:
+        for v in chain(self.utility_visitors, self.visitors):
```
Zac-HD, line 128:
```suggestion
        for v in (*self.utility_visitors, *self.visitors):
```

Could also use `+` because they're both tuples, but the "anything iterable" syntax is visibly safe.

CONTRIBUTING.md
jakkdl: I like the generated files for checking that they're valid and passes linters, and when developing they're easier to read and throw into `cst.parse_module`, but I'll generate diff files as well :+1:

Zac-HD: Oh, yep, both sounds good!

jakkdl: Wohoo, bigger commits! :grin:
They're still not perfect, as they're not showing e.g. where checkpoints *aren't* inserted - but that will be covered by future tests that autofixed files don't generate any errors. (and will have to figure out a solution for errors that are expected not to get autofixed)

flake8_trio/runner.py
jakkdl: Thought the `chain` was nice and explicit, but don't mind either variant. The plus solution might also be slower, no?

jakkdl:
fixed comments

jakkdl:
except I had untracked files ... updated my aliases so that won't happen again ^^
This is so handy it's crazy:
```sh
alias untracked="git ls-files -o --directory --exclude-standard | sed q1 > /dev/null"
alias gfpush='untracked && pytest --quiet -x && git commit -a --amend --no-edit && git push -f'
```

jakkdl:
Oh and I realized that the `pass # AUTOFIX_INSERT` was a big brainfart yesterday, autofixing is *all about* inserting lines - I'm not explicitly marking each checkpoint that should be inserted, they're just added/generated to the autofix file. So completely removed that.
Still need the `isort: skip_file` though since autofix doesn't insert an empty line before the added import, and trying to worry about that isn't really viable.

jakkdl:
ahahaha, `trio.lowlevel.checkpoint()` sure doesn't checkpoint - you kiiiinda need to do `await trio.lowlevel.checkpoint()`

jakkdl:
I'm writing the extra tests now, and this implementation still does some dumb stuff (yield inside boolops and stuff), but I'll leave (not) handling those for the next PR.

Zac-HD:
> `alias gfpush='untracked && pytest --quiet -x && git commit -a --amend --no-edit && git push -f'`

`--force-with-lease`!

> ahahaha, `trio.lowlevel.checkpoint()` sure doesn't checkpoint - you kiiiinda need to do `await trio.lowlevel.checkpoint()`

lint all the things 😜

jakkdl:
Ooh, `--force-with-lease` is nice!

Well, there's no linter - other than flake8-trio itself, that catches unawaited `trio.lowlevel.checkpoint()`! Unless we change the pyright dependency from `trio` to `trio-typing`. That might definitely be good though, but I'm getting 59 errors when doing that - so needs a separate PR.

Zac-HD (approved):
Whew!  Big diff to review, but this looks great.  Thanks again @jakkdl!

jakkdl:
I'll try to get the diffs to be smaller, but it's hard to avoid at this stage - and especially when working with the monster that is 91x :sweat_smile:

Zac-HD:
Oh yeah, that's not a criticism! Just an expression of relief that I made it through 😋

jakkdl:
Haha, just meant it's worth trying for. If I don't keep it in the back of my mind the scope creep of PR's is very real :innocent:

PR #161, opened 2023-03-21
add libcst as install_requires, print errors to stderr & return 1/0, add pre-commit hook & update config, 

jakkdl:
Didn't come up with any good ways of writing an automated test for pre-commit, but tested it locally and against my fork - and therefore finding the catastrophically bad errors in init & setup :sweat_smile: 
I was a bit surprised missing libcst wasn't picked up by tox, but that's because the testenv depends on hypothesmith - which requires libcst.

We surely want to start tagging releases though, I looked at shed's CI workflow if there were anything there that automatically created tags but didn't find anything so I'm a bit curious if/how you automate that or if you manually do it on releasing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,7 @@ repos:
           - id: black
 
     - repo: https://github.com/PyCQA/autoflake
-      rev: v2.0.1
+      rev: v2.0.2
       hooks:
           - id: autoflake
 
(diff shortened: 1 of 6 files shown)


jakkdl:
rejigged the tests a bit to have two tests explicitly for SystemExit.

Zac-HD (approved):

Zac-HD:
> I looked at shed's CI workflow if there were anything there that automatically created tags but didn't find anything so I'm a bit curious if/how you automate that or if you manually do it on releasing.

Automated via [this ci config](https://github.com/Zac-HD/shed/blob/a3853ea5db56e28806035376de19e289681aff9e/.github/workflows/ci.yml#L68) invoking [this janky code](https://github.com/Zac-HD/shed/blob/a3853ea5db56e28806035376de19e289681aff9e/tests/test_version.py#L51-L80)!  It's not pretty, but it works 😅

PR #162, opened 2023-03-21
Run black with --preview

jakkdl:
They've improved multi-line string handling, which I otherwise find quite irritating with the loads of long strings everywhere (e.g. currently writing help texts for command line args).

### DIFF
--- a/tests/test_flake8_trio.py
+++ b/tests/test_flake8_trio.py
@@ -293,9 +293,7 @@ def _parse_eval_file(test: str, content: str) -> tuple[list[Error], list[str]]:
             try:
                 expected.append(Error(err_code, lineno, int(col), message, *args))
             except AttributeError as e:
-                msg = (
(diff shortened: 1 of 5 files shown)


tests/test_flake8_trio.py
```diff
@@ -293,9 +293,7 @@ def _parse_eval_file(test: str, content: str) -> tuple[list[Error], list[str]]:
             try:
                 expected.append(Error(err_code, lineno, int(col), message, *args))
             except AttributeError as e:
-                msg = (
-                    f"Line {lineno}: Failed to format\n {message!r}\n" f'"with\n{args}'
-                )
+                msg = f'Line {lineno}: Failed to format\n {message!r}\n"with\n{args}'
```
Zac-HD, line 296: The internal `"` mark seems strange to me, but happy to revisit later if you want to.

Zac-HD (approved):

PR #159, opened 2023-03-17
test for autofix_files, fix some 91x autofixes, rejig eval_files

jakkdl:
branched on top of #158 
F_ixes the sixth checkbox in #124 - "test that autofixed files don't generate errors", and also tests that autofixed files don't get further autofixes applied if run through the transformer again.

~~Finished, but somewhat messy - want to take a fresh look at `visitor91x.py` and the class structure there another day to see if I can make it less awful.~~

eval/autofix diffs are quite irritating when line numbers change, both in showing up in the diff and when developing when I manually edit the files - so it's possible I tweak something there in the future.

~~Feel free to review if you want, but I plan on cleaning it up and adding review comments so it should be easier after that.~~

PR #163, opened 2023-03-21
add standalone command line args

jakkdl:
And just for funsies let's open a fourth PR :grin: 
It's branched on top of #162, so the diff will be *slightly* smaller once that's merged.

Ready for review, though I can probably polish it a bit if giving it another look tomorrow.
Had a decently long thonk on whether to just have an `--enable` and no `--disable`, or whether to get the full flake8 experience with `--extend-[...]` flags. Settled on this compromise.

91x does not respect `--autofix` parameter value properly atm, hence the commented out code with that, would like #159 in order to work on that.

Will probably have some merge conflicts and need rebasing once other PR's get merged

### DIFF
--- a/flake8_trio/__init__.py
+++ b/flake8_trio/__init__.py
@@ -15,7 +15,6 @@ import ast
 import functools
 import keyword
 import os
-import re
 import subprocess
 import sys
 import tokenize
@@ -24,8 +23,9 @@ from typing import TYPE_CHECKING
 
 import libcst as cst
 
+from .base import Options
 from .runner import Flake8TrioRunner, Flake8TrioRunner_cst
-from .visitors import default_disabled_error_codes
+from .visitors import ERROR_CLASSES, ERROR_CLASSES_CST, default_disabled_error_codes
 
 if TYPE_CHECKING:
     from collections.abc import Iterable, Sequence
@@ -76,12 +76,6 @@ def cst_parse_module_native(source: str) -> cst.Module:
 
 def main() -> int:
     parser = ArgumentParser(prog="flake8_trio")
-    parser.add_argument(
-        nargs="*",
-        metavar="file",
-        dest="files",
-        help="Files(s) to format, instead of autodetection.",
-    )
     Plugin.add_options(parser)
     args = parser.parse_args()
     Plugin.parse_options(args)
@@ -115,7 +109,7 @@ def main() -> int:
         for error in sorted(plugin.run()):
             print(f"{file}:{error}")
             any_error = True
-        if plugin.options.autofix:
+        if plugin.options.autofix_codes:
             with open(file, "w") as file:
                 file.write(plugin.module.code)
     return 1 if any_error else 0
@@ -124,7 +118,13 @@ def main() -> int:
 class Plugin:
     name = __name__
     version = __version__
-    options: Namespace = Namespace()
+    standalone = True
+    _options: Options | None = None
+
+    @property
+    def options(self) -> Options:
+        assert self._options is not None
+        return self._options
 
     def __init__(self, tree: ast.AST, lines: Sequence[str]):
         super().__init__()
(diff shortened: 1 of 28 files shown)


flake8_trio/__init__.py
```diff
@@ -124,7 +118,13 @@ def main() -> int:
...
     name = __name__
     version = __version__
-    options: Namespace = Namespace()
+    standalone = True
+    _options: Options | None = None
+
+    @property
+    def options(self) -> Options:
+        assert self._options is not None
+        return self._options
```
jakkdl, line 127: Options is a dataclass, so we get type checking on options.

flake8_trio/__init__.py
```diff
@@ -208,19 +254,6 @@ def add_options(option_manager: OptionManager | ArgumentParser):
...
-            default=".*",
-            required=False,
-            help=(
-                "Regex string of visitors to enable. Can be used to disable broken "
-                "visitors, or instead of --select/--disable to select error codes "
-                "in a way that is more performant. If a visitor raises multiple codes "
-                "it will not be disabled unless all codes are disabled, but it will "
-                "not report codes matching this regex."
-            ),
-        )
```
jakkdl, line 223: use `--enable` instead

flake8_trio/base.py
```diff
@@ -2,7 +2,24 @@
...
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Any, NamedTuple
+
+if TYPE_CHECKING:
+    from collections.abc import Collection
+
+
+@dataclass
+class Options:
+    # enable and disable have been expanded to contain full codes, and have no overlap
```
jakkdl, line 14: not sure having both enable and disable is any upside. And if anything should have a enabled_or_autofixed ?

flake8_trio/runner.py
```diff
@@ -21,44 +21,48 @@
...
+    options: Options
     problems: list[Error] = field(default_factory=list)
     library: tuple[str, ...] = ()
     typed_calls: dict[str, str] = field(default_factory=dict)
     variables: dict[str, str] = field(default_factory=dict)
 
 
-class Flake8TrioRunner(ast.NodeVisitor):
-    def __init__(self, options: Namespace):
+class CommonRunner(ABC):  # noqa: B024 # no abstract methods
```
jakkdl, line 41: split out so `selected` doesn't have to be updated in two places

flake8_trio/visitors/helpers.py
```diff
@@ -114,10 +114,8 @@ def fnmatch_qualified_name_cst(
 # used in 103/104
 def iter_guaranteed_once(iterable: ast.expr) -> bool:
     # static container with an "elts" attribute
-    if hasattr(iterable, "elts"):
-        elts: Iterable[ast.expr] = iterable.elts  # type: ignore
-        for elt in elts:
-            assert isinstance(elt, ast.expr)
+    if isinstance(iterable, (ast.List, ast.Tuple, ast.Set)):
+        for elt in iterable.elts:
```
jakkdl, line 118: rejigged to avoid `type: ignore`

flake8_trio/visitors/helpers.py
```diff
@@ -354,16 +352,14 @@ def func_has_decorator(func: cst.FunctionDef, *names: str) -> bool:
...
+        )
+    else:
+        for n in node:
             yield from get_comments(n)
-        return
-    yield from (
-        cst.EmptyLine(comment=ensure_type(c, cst.Comment))
-        for c in m.findall(cast("cst.CSTNode", node), m.Comment())
-    )
-    return
```
jakkdl, line 366: pyright updated and started requiring even *more* casts (or if that was elsewhere, idr), so rewritten to not require any casts. Although the MaybeSentinel not being in the function header is ... not great

flake8_trio/runner.py
```diff
@@ -21,44 +21,48 @@
...
+    def __init__(self, options: Options):
         super().__init__()
         self.state = SharedState(options)
 
+    def selected(self, error_codes: dict[str, str]) -> bool:
+        for code in error_codes:
+            for enabled in self.state.options.enable, self.state.options.autofix:
+                if code in enabled:
+                    return True
+        return False
```
Zac-HD, line 51: I find it easier to think about this in terms of set operations:

```suggestion
        enabled_codes = set(self.state.options.enable) | set(self.state.options.autofix)
        return bool(enabled_codes & set(error_codes))
```

tests/test_flake8_trio.py
```diff
@@ -48,14 +47,6 @@ class ParseError(Exception):
...
 
 
-# flake8 6 added a required named parameter formatter_names
-def _default_option_manager():
-    kwargs = {}
-    if flake8_version_info[0] >= 6:
-        kwargs["formatter_names"] = ["default"]
-    return OptionManager(version="", plugin_versions="", parents=[], **kwargs)
-
-
```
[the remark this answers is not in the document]
Zac-HD: My usual approach here is to keep the extra tests until the first time you find them annoying, and then delete them unless they've _also_ been useful by then.  Maximally lazy development!

tests/test_flake8_trio.py
```diff
@@ -319,13 +317,14 @@ def _parse_eval_file(test: str, content: str) -> tuple[list[Error], list[str]]:
                 msg = f'Line {lineno}: Failed to format\n {message!r}\n"with\n{args}'
                 raise ParseError(msg) from e
 
-    assert visitor_codes_regex, "no visitors enabled"
     for error in expected:
-        assert re.match(
-            visitor_codes_regex, error.code
-        ), "Expected error code not enabled"
+        for code in enable_codes.split(","):
+            if error.code.startswith(code):
```
Zac-HD, line 322: I'm concerned that a repeated-comma typo would enable everything here, might be worth checking that the code starts with `TRIO`?

flake8_trio/visitors/flake8triovisitor.py
```diff
@@ -228,6 +224,12 @@ def error(
             )
         )
 
+    def autofix(self, code: str | None = None):
```
Zac-HD, line 227: Maybe `should_autofix`?  This seems more like an imperative than a logical test.

Zac-HD (approved):
Looks good to me; some comments below but feel free to merge when you've addressed them (or decided not to) to your satisfaction 👍

tests/test_flake8_trio.py
jakkdl: got confused for a moment that this was talking about the user-supplied string, but added tests that it won't break for users, and improved logic in the test so it'll warn about bad codes.

flake8_trio/visitors/flake8triovisitor.py
jakkdl: for sure, changed

tests/test_flake8_trio.py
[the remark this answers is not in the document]
jakkdl: it doubles runtime of tests in CI, which definitely is quite annoying. I guess one could rejig the CI to run flake8_5 separately, but eh...

flake8_trio/base.py
jakkdl: ye I realized as I wrote my comment that it was bad design, also renamed the variables and improved comments.

PR #165, opened 2023-03-22
run fuzz in CI

jakkdl:
let's see how long they take to run in CI, I usually don't run them in my workflow locally since they're slow - but it's probably nice if they're run regularly *somewhere*

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,23 @@ jobs:
             - name: Run tests
               run: python -m tox -e flake8_5,flake8_6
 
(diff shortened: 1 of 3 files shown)


jakkdl:
The check is once again fixed in #163, I kind of expect the fuzz tests to fail currently - so make sure to rerun after merging the others ... if the runtime looks the least bit reasonable.

Zac-HD:
> with "non-required" do you simple mean not listing it in `needs` for the release test?

I mean "oops, let me go set up some branch protection rules so you can't merge a PR until the tests pass" 😆

PR #167, opened 2023-03-23
clean up tests, merging test_eval and test_eval_anyio

jakkdl:
somewhat messy diff since one test function is deleted and the other is modified, but overall functionality should be the exact same.

somewhat awkward to type hint with repeating the literals twice, feels like it should be possible to get rid of that.

### DIFF
--- a/tests/test_flake8_trio.py
+++ b/tests/test_flake8_trio.py
@@ -15,7 +15,7 @@ import unittest
 from argparse import ArgumentParser
 from collections import deque
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, DefaultDict
+from typing import TYPE_CHECKING, Any, DefaultDict, Literal
 
 import libcst as cst
 import pytest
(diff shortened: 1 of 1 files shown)


tests/test_flake8_trio.py
```diff
@@ -157,70 +157,68 @@ def check_autofix(
...
 
 
-# TODO: merge with test_eval_anyio with a parametrize
-@pytest.mark.parametrize(("test", "path"), test_files)
-@pytest.mark.parametrize("autofix", [False, True])
-def test_eval(test: str, path: Path, autofix: bool, generate_autofix: bool):
-    content = path.read_text()
-    if "# NOTRIO" in content:
-        pytest.skip("file marked with NOTRIO")
+MAGIC_MARKERS = ("NOANYIO", "NOTRIO", "TRIO_NO_ERROR", "ANYIO_NO_ERROR")
```
Zac-HD, line 160: > somewhat awkward to type hint with repeating the literals twice, feels like it should be possible to get rid of that.

We could define `MAGIC_MARKERS: TypeAlias = Literal[...]` here, and then use `MAGIC_MARKERS.__args__` where you iterate the strings?

Zac-HD (approved):

PR #169, opened 2023-03-24
Clean up typechecking for MAGIC_MARKERS by making it a dataclass, add…

jakkdl:
… comments w/ possible further improvements

Followup after #167

### DIFF
--- a/tests/test_flake8_trio.py
+++ b/tests/test_flake8_trio.py
@@ -14,8 +14,9 @@ import tokenize
 import unittest
 from argparse import ArgumentParser
 from collections import deque
+from dataclasses import dataclass, fields
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, DefaultDict, Literal
+from typing import TYPE_CHECKING, Any, DefaultDict
 
 import libcst as cst
 import pytest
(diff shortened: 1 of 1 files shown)


tests/test_flake8_trio.py
```diff
@@ -157,18 +158,30 @@ def check_autofix(
...
+    found_markers = MagicMarkers()
+    markers = (f.name for f in fields(found_markers))
+    pattern = rf'# ({"|".join(markers)})'
+    for f in re.findall(pattern, content):
+        setattr(found_markers, f, True)
+    return found_markers
 
 
+# This could be optimized not to reopen+reread+reparse the same file over and over
+# when testing the same file
```
Zac-HD, line 184: Turn `find_magic_markers()` into a classmethod on the dataclass, with `@lru_cache`?

Zac-HD (approved):
Merging as improvement on status quo 👍

tests/test_flake8_trio.py
jakkdl: it's not just find_magic_markers, but also `_parse_eval_file`. The latter would need some additional logic since the errors & error messages will be slightly different when substituting trio for anyio.

PR #179, opened 2023-04-21
Fix false 91x alarm on await in genexp target.

### LINKED ISSUE #174: False alarms in `libcst` backend
```python
import trio

# Seems like the inner context manager 'hides' the checkpoint.
async def does_contain_checkpoints():
    with trio.fail_after(1):  # false-alarm TRIO100
        with trio.CancelScope():  # or any other context manager
            await trio.sleep_forever()

# -------------------------

async def elems():
    print("here")
    return ["elem1", "elem2"]

async def await_in_comp_target():
    return (print(x) for x in await elems())  # false-alarm TRIO910

trio.run(await_in_comp_target)  # prints "here", although not any elements
```

jakkdl:
Fixes #174

### DIFF
--- a/flake8_trio/visitors/visitor91x.py
+++ b/flake8_trio/visitors/visitor91x.py
@@ -769,9 +769,11 @@ class Visitor91X(Flake8TrioVisitor_cst, CommonVisitors):
 
         return False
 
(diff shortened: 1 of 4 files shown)


Zac-HD (approved):
Sweet!

PR #182, opened 2023-04-28
respect noqa

### LINKED ISSUE #181: Support `# noqa` comments when calling directly
When I tried calling `flake8_trio` directly from the commandline, it turns out that doesn't respect `# noqa` comments - which would unfortunately block direct use for now.  If supporting them is viable, let's do that and add a `--disable-noqa` flag for the current behavior.

I'd also prefer to name the cli tool `flake8-trio`; underscores are pretty unusual in the terminal.

Once this is in, I think we're ready to release the big refactoring!

jakkdl:
fixes #181 
Implemented:
* respects `noqa`
* new pyright error about how to in-line silence stuff.
  * it would be nice to catch errors-due-to-updated-linters with a bot, instead of me randomly encountering them when writing new PRs and being surprised.
* various small test improvements
  * changes parametrized params to `test_eval` so summaries are more informative than e.g. `test_eval[False, True, TRIO100, path7]`
  * sort errors on columns late, since columns have a habit of getting messed up and thrown around (and might be ignored anyway)
  * ignore columns when checking with autofix
    * the scenario here is two errors on the same line, the first which modifies the line, causing the second to have a different column

remaining TODO:
* don't autofix `noqa`'d errors
  * ... ah shit, I have to kind of completely change when errors get suppressed to fix this - since visitors have to know if an error is `noqa`d when they raise/autofix it.
* support `noqa-trio` (or similar / depends on discussion in #181)
  * handle NQA102 for `noqa-trio`
    * separate error, or command-line flag, if code not enabled
  * check that ignored code is valid/exists
    * maybe give separate error if it's not enabled, see above
* autofixing two errors in the same line seems broken
  * TRIO100 autofix looks like it's eating comments
  * TRIO911 autofix does not get executed, although that's possibly due to it being a `SimpleStatementSuite`
  * might fix these in a separate PR which i rebase this onto, or just do `NOAUTOFIX`
* noqas with invalid code is parsed as bare noqa (but that's same as flake8, so /shrug )
* report old regex101.com link upstream in PyCQA/flake8
* don't suppress errors when running as a plugin
* `--disable-noqa`

### DIFF
--- a/tests/test_flake8_trio.py
+++ b/tests/test_flake8_trio.py
@@ -108,7 +108,13 @@ def check_autofix(
     visited_code = plugin.module.code
 
     if "# AUTOFIX" not in unfixed_code:
-        assert unfixed_code == visited_code
+        # if the file is specifically marked with NOAUTOFIX, that means it has visitors
+        # that will autofix with --autofix, but the file explicitly doesn't want to check
+        # the result of doing that. THIS IS DANGEROUS
+        if "# NOAUTOFIX" in unfixed_code:
+            print(f"eval file {test} marked with dangerous marker NOAUTOFIX")
+        else:
+            assert unfixed_code == visited_code
         return
 
     # the full generated source code, saved from a previous run
@@ -185,9 +191,9 @@ def find_magic_markers(
 
 # This could be optimized not to reopen+reread+reparse the same file over and over
 # when testing the same file
-@pytest.mark.parametrize(("test", "path"), test_files)
-@pytest.mark.parametrize("autofix", [False, True])
-@pytest.mark.parametrize("anyio", [False, True])
+@pytest.mark.parametrize(("test", "path"), test_files, ids=[f[0] for f in test_files])
+@pytest.mark.parametrize("autofix", [False, True], ids=["noautofix", "autofix"])
+@pytest.mark.parametrize("anyio", [False, True], ids=["trio", "anyio"])
 def test_eval(
     test: str, path: Path, autofix: bool, anyio: bool, generate_autofix: bool
 ):
@@ -196,7 +202,9 @@ def test_eval(
     if anyio and magic_markers.NOANYIO:
         pytest.skip("file marked with NOANYIO")
 
-    ignore_column = False
+    # if autofixing, columns may get messed up
+    ignore_column = autofix
+
     if magic_markers.NOTRIO:
         if not anyio:
             pytest.skip("file marked with NOTRIO")
@@ -241,6 +249,8 @@ def test_eval(
 @pytest.mark.parametrize("test", autofix_files)
 def test_autofix(test: str):
     content = autofix_files[test].read_text()
+    assert content, "empty file"
+
     if "# NOTRIO" in content:
         pytest.skip("file marked with NOTRIO")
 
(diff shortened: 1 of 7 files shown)


Zac-HD:
> it would be nice to catch errors-due-to-updated-linters with a bot, instead of me randomly encountering them when writing new PRs and being surprised.

Yeah, I think it's time that we started pinning our dependencies.  Steal the `pip-tools` setup from `shed`?  The idea is to pin all the transitive deps, and then have a tox env that updates them - for now we can just run that manually every so often.

Zac-HD:
> don't autofix `noqa`'d errors

For simplicity I think we can skip this too for now; if it's autofixable I think we're OK with doing that and if I turn out to be wrong, well, that's a future problem 😝

tests/test_flake8_trio.py
```diff
@@ -186,17 +186,24 @@ def find_magic_markers(
...
+@pytest.mark.parametrize("autofix_str", ["noautofix", "autofix"])
+@pytest.mark.parametrize("library", ["trio", "anyio"])
 def test_eval(
-    test: str, path: Path, autofix: bool, anyio: bool, generate_autofix: bool
+    test: str, path: Path, autofix_str: str, library: str, generate_autofix: bool
 ):
+    # pytest prints parametrized values, but not the key, so to make the summary info
+    # readable we make the parameters strings and here turn them into bools
+    autofix = autofix_str == "autofix"
+    anyio = library == "anyio"
```
Zac-HD, line 197: There's also the `ids=` argument which can customize this, passing either a list of strings or an arg->str function.

tox.ini
```diff
@@ -53,6 +53,9 @@ extend-enable = TC10
 exclude = .*, tests/eval_files/*, tests/autofix_files/*
 per-file-ignores =
     flake8_trio/visitors/__init__.py: F401, E402
+# visitor_utility contains comments specifying how it parses noqa comments, which get
+# parsed as noqa comments
+    flake8_trio/visitors/visitor_utility.py: NQA101, NQA102
```
Zac-HD, line 58: 🤣

Zac-HD (commented):
looks good!

tests/test_flake8_trio.py
jakkdl: ooh, much better!

jakkdl:
> > don't autofix `noqa`'d errors
> 
> For simplicity I think we can skip this too for now; if it's autofixable I think we're OK with doing that and if I turn out to be wrong, well, that's a future problem stuck_out_tongue_closed_eyes

well, if you have any TRIO91x errors noqa'd they'll be autofixed - although I suppose extra lowlevel checkpoints won't do anything. noqa'd TRIO100 being autofixed is quite a bit worse, but maybe not too common. But I can postpone it to a separate PR regardless.

jakkdl:
Oh actually, current solution is very incompatible with flake8-noqa as it suppresses errors even when run non-standalone...

jakkdl:
But feel free to merge and test against live code, and if it works up to you how badly you want to break flake8-noqa, or wait until followup PRs.

jakkdl:
nowait, this also needs to ignore ast visitor errors

Zac-HD (approved):

Zac-HD:
Looks like we've got some useful fixes here for the other PRs, so merging this + #184 now.  I'll test locally ~monday (I hope 🤞), and cut a release either then or after #185 if that turns out to be really necessary.  Thanks again!

PR #183, opened 2023-04-28
new error code: TRIO912 unnecessary checkpoints + autofix

### LINKED ISSUE #70: Idea: warn on unnecessary checkpoints added to avoid TRIO107/108

jakkdl:
fixes #70 (merely 5 months after it was opened :grin:)

Early WIP, I only got the basic case down.
I started on this before #182, but I'm prioritizing that one.

Remaining TODO:
* `if` statements
  * this shouldn't be *that* complicated, I've mostly got the logic figured out.
* loops
  * once `if` is solved this can probably use similar logic, but it'll likely be messy
* try/except
  * same

### DIFF
--- a/flake8_trio/visitors/visitor91x.py
+++ b/flake8_trio/visitors/visitor91x.py
@@ -56,6 +56,9 @@ class LoopState:
     nodes_needing_checkpoints: list[cst.Return | cst.Yield] = field(
         default_factory=list
     )
+    possibly_redundant_lowlevel_checkpoints: list[cst.BaseExpression] = field(
+        default_factory=list
+    )
 
     def copy(self):
         return LoopState(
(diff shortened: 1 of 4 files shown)


jakkdl:
I picked this up and worked on it somewhat recently, but stalled out due to the complexity. The complexity is mostly because it's hard to resist solving the juicy edge cases and autofixing everything, so next time I pick this up I'll just drop the complex parts, see if the basic cases that are working are worth having as a rule, and clean up the PR so it can be merged.

PR #186, opened 2023-04-30
fix trio100 removing trailing comment with same-line with

jakkdl:
added basic test case showing the problem, this was the cause of my confusion in #182 with comments disappearing.

todo: write a fix


jakkdl:
thought it'd be trickier than this, but past me had written good code and libcst is nice :relaxed:

Zac-HD (approved):
Wow, that's even nicer than I expected from your comment 🤩

Issue #185, opened 2023-04-30
full `# noqa` support


jakkdl:
Making a separate issue from #181 to track all current issues/TOODs with `# noqa`

* [ ] Don't autofix `noqa`'d errors
  * This might not be a huge issue atm, but is a must going forward as more visitors get autofix support.
* ~~Respect `noqa-trio` as a way to have full flake8-noqa support~~
* [ ] Raise error if invalid (or not enabled) TRIO error code is noqa'd.
  * Not a must, and sort of covered by `flake8-noqa`. But if it's a simple logic addition there's not much reason not to.
* [ ] raise `noqa`d errors when running as a plugin
  * otherwise we're fully incompatible with flake8-noqa
* [ ] support `--disable-noqa`
* [ ] document current level of [in]compatibility with flake8-noqa

Issue #187, opened 2023-04-30
pin test versions in pre-commit


jakkdl:
I have not evaluated alternatives or looked at shed's pip-tools setup https://github.com/Zac-HD/flake8-trio/pull/182#issuecomment-1528387080, just making an issue to track it for now

jakkdl:
By default versions *are* pinned in pre-commit, but I had set `env PYRIGHT_PYTHON_FORCE_VERSION=latest` - probably just because I was irritated with the warning that there was a new version available (turns out there's a `PYRIGHT_PYTHON_IGNORE_WARNINGS` i can use :sweat_smile:). So the immediate problem should be solvable merely by not setting that environment variable - but I would like to setup auto-updates now that development has slowed down so stuff like updates to flake8 doesn't catch us off guard again. And auto-updating the linter versions in pre-commit is nice too.


shed's `pip-tools` setup doesn't integrate with pre-commit, and the `requirements.[txt/ini]` file in trio has made me gone mad - though my frustrations with it could probably be addressed by running it from tox with a specified python version (which I notice shed doesn't, so `tox -e deps` on non-python3.11 will lead to changes).

#### For pre-commit, options include:
* pre-commit.ci bot
  * pro: CI checks being faster as it caches versions.
  * con: to see output you have to go to pre-commit's site (I don't personally experience this as much of a con, but you mentioned it previously)
  * While pyright can't run inside the pre-commit CI bot (since that lacks internet access, we currently trigger it manually from `ci.yml`), it'll still see version updates and trigger a PR - in which the GH action will run pyright as usual
* GH action that runs pre-commit autoupdate https://browniebroke.com/blog/gh-action-pre-commit-autoupdate/
* more esoteric home-grown solutions.

I'd personally like to try out pre-commit.ci to see how that works. So for that you need to sign in with github and add it at https://pre-commit.ci/
 

#### For direct dependencies
it's either dependabot or custom GH actions. This likely will be triggered much less often with only two dependencies that can update, and seeing dependabot everywhere I'd like to familiarize myself with it. Setting it up should be fairly trivial: https://til.simonwillison.net/github/dependabot-python-setup - I can push a PR with the yaml file, but after that you'd need to push the buttons in the settings.


#### pyright GH-action
[...]

Issue #160, opened 2023-03-20
Change trio dependency for pyright to trio-typing


jakkdl:
Since there's a bunch of stuff in `trio` that doesn't raise errors that `trio-typing` does, this would raise a bunch of warnings about eval files. But as typing in `trio` converges to the bar in `trio-typing` these errors would have to be addressed at some point sooner or later anyway.
Currently raises 60-ish errors, which mostly will just be ignored. (e.g. 35 errors in TRIO105 about expressions not being awaited ... because that's what the check tests for)

jakkdl:
trio-typing is now deprecated as typehints have been added to mainline trio.

PR #190, opened 2023-05-02
don't autofix noqa'd errors, add --disable-noqa, add non-passing test for noqa reliability

jakkdl:
big refactor for noqa functionality fixing most of the remaining ones in #185, but as I added the test for all errors being properly noqa'd and was greeted with a wall of errors I realized it might be good to commit and push before tackling them. If it's a minor fix I'll just push/amend it to this PR, otherwise open a separate one.

* ~~Moves `# noqa` logic into `Flake8TrioVisitor_cst`, other than for ast visitors where some logic still is in `__init__.py`~~
  * ~~utility visitor `NoqaHandler` removed~~
* add `--disable-noqa`, and use that instead of whether we're standalone to decide whether to parse noqa or not.
* ~~the list of `noqas` is no longer parsed once by a utility visitor, but instead parsed by each visitor. This is a performance hit (though probably not very large, if any) but is sorta necessary as autofixing messes up line numbers.~~ autofixing does *not* mess up line numbers in fact.
* `should_autofix` now takes a `CSTNode` parameter so it can determine the line number of it, and check if it's noqa'd
* ~~`visitor91x` now doesn't autofix simple `noqa` scenarios, but doesn't handle loops and the like~~ visitor91x handles all cases
* tests and shit

### DIFF
--- a/flake8_trio/__init__.py
+++ b/flake8_trio/__init__.py
@@ -150,21 +150,24 @@ class Plugin:
         return plugin
 
     def run(self) -> Iterable[Error]:
-        problems_ast = Flake8TrioRunner.run(self._tree, self.options)
-
-        cst_runner = Flake8TrioRunner_cst(self.options, self.module)
-        problems_cst = cst_runner.run()
-
         # when run as a flake8 plugin, flake8 handles suppressing errors from `noqa`.
         # it's therefore important we don't suppress any errors for compatibility with
         # flake8-noqa
         if not self.standalone:
+            self.options.disable_noqa = True
+
+        cst_runner = Flake8TrioRunner_cst(self.options, self.module)
+        # any noqa'd errors are suppressed upon being generated
+        yield from cst_runner.run()
+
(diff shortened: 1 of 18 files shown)


jakkdl:
turns out fixing the errors made the PR simpler, as I had mistaken assumptions about how lineno-parsing was handled in `CSTTransformers`.

jakkdl:
uh, is it an update to trio or pyright that has it throwing new errors? It now seems to be analyzing `trio` properly for some reason and seeing missing objects and whether functions are coroutines.

Zac-HD:
I don't know why `pyright` has changed, but short-term I'd just ignore `tests/` and longer term #187.

jakkdl:
> I don't know why `pyright` has changed, but short-term I'd just ignore `tests/` and longer term #187.

added commit which excludes `tests/eval_files` and `tests/autofix_files`

jakkdl:
> I don't know why `pyright` has changed, but short-term I'd just ignore `tests/` and longer term #187.

also related to #160

PR #196, opened 2023-05-22
replace flake8-eradicate with ruff ... and enable a ton of more lint rules, fixing a couple of them.

jakkdl:
If enabling auto-updating of pre-commit, flake8-eradicate will come to bite us as it's not accepting PR's to show that it works with flake8 6. I was planning to add an updated fork as additional_dependency to flake8 instead, but I stumbled upon https://github.com/charliermarsh/ruff instead - which, among others, reimplements flake8-eradicate. Curiosity peaked, I added it to CI, enabled *everything* (it sucks there's no `--select-all`), and after playing around for a bit it looks to be almost strictly better than flake8 in a lot of ways. It implements a ton of plugins by default, has autofix for a bunch of them, and in general very easy to transition to from flake8.

flake8-mutable is the only plugin flake8-trio uses I didn't see listed in https://github.com/charliermarsh/ruff#rules (though maybe there is a rule for default mutable arguments anyway, haven't checked), and flake8-noqa (but there is support for checking noqa stuff).
Looks like they haven't kept up reimplementing *all* rules from flake8-bugbear (esp not from the opinionated ones), but majority of them are in.

So might be interesting to enable it, and see how viable it is as a replacement for flake8.


flake8_trio/__init__.py
```diff
@@ -351,7 +351,8 @@ def parse_trio114_identifiers(raw_value: str) -> list[str]:
     values = comma_separated_list(raw_value)
     for value in values:
         if keyword.iskeyword(value) or not value.isidentifier():
-            raise ArgumentTypeError(f"{value!r} is not a valid method identifier")
+            msg = f"{value!r} is not a valid method identifier"
+            raise ArgumentTypeError(msg)
```
Zac-HD, line 355: I prefer the old style here, can we disable that lint and revert?

flake8_trio/visitors/flake8triovisitor.py
```diff
@@ -9,12 +9,12 @@
 import libcst as cst
 from libcst.metadata import PositionProvider
 
-from ..base import Error, Statement
+from flake8_trio.base import Error, Statement
```
Zac-HD, line 12: Similarly I think local imports are fine, let's disable + revert that too.

flake8_trio/visitors/visitor100.py
```diff
@@ -7,9 +7,8 @@
 """
 from __future__ import annotations
 
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
-import libcst as cst
```
Zac-HD, line 12: Aaaand, let's disable this as well - it's pretty verbose and doesn't actually have any performance benefits here at all; we're executing all the imports already when we import the matchers.

(perils of starting with such a long list of available lint rules!)

Zac-HD (commented):
Nice, I've been vaguely interested in trying `ruff` for a while and this looks like a clean replacement.  IMO we could drop the flake8 checks immediately, if there's nothing else left out?

flake8_trio/visitors/visitor100.py
jakkdl: Huh, looks like flake8-type-checking has explicit logic about ignoring the case when you import a submodule. I'll add `libcst` to https://beta.ruff.rs/docs/settings/#exempt-modules for now but might want to stick with `flake8-type-checking` over ruff and/or open an issue in ruff.

jakkdl:
> IMO we could drop the flake8 checks immediately, if there's nothing else left out?

I don't mind running both for a while to see where they agree/disagree. But if it becomes bothersome I'll quickly start disabling error codes/plugins, set the [`stage`](https://pre-commit.com/#confining-hooks-to-run-at-certain-stages) for flake8 to `manual`, and/or remove it.

flake8_trio/visitors/visitor100.py
jakkdl: opened https://github.com/charliermarsh/ruff/issues/4667

Zac-HD (approved):
:shipit:

PR #202, opened 2024-02-15
Update ci

jakkdl:
* bump pre-commit checkers
* use actions/checkout@v4
* use actions/setup-python@v5
* New Black style guide
* fix mypy errors
  * `if ...:` is now treated as `if True:` and it therefore thought tons of the `eval_files` contained unreachable statements. Replaced it with an arbitrary function call.
* fix pyright errors
  * ignored error about `Statement` being unhashable, [which it technically should be](https://docs.python.org/3/reference/datamodel.html#object.__hash__), although I guess we're inheriting a `__hash__` from `NamedTuple` and it currently works :tm:. Need to change a bunch of equality checks in the code to make it less illegal, so leaving that for now.
  * https://github.com/Instagram/LibCST/issues/1107
* ruff
  * bug in TCH002 fixed: https://github.com/astral-sh/ruff/issues/4667
  * 2.0 has new recommended config setup
* remove flake8 from pre-commit (it was only kept for curiosity, and started giving new errors)
* add py312 in CI
* pyupgrade broke a test file and gave invalid code, so that file is now excluded
  * https://github.com/asottile/pyupgrade/issues/932
* TRIO910 "non-empty async function must contain checkpoint" did not properly detect
```python
async def foo(): ...
```
as an empty function, since it was matching against the indented body. This is now the black style, which is why I found it. So fixed that, and updated test file to have both.
There's only a few more instances of `IndentedBlock` in the project, and none seem problematic, but it's possible this could be a problem in other places.
* adding 3.13 to CI is blocked by https://github.com/Zac-HD/hypothesmith/issues/36

I can split this across multiple PRs, but idk if it's warranted. Now that we moved and can set up pre-commit it should be more incremental going forward.

### DIFF
--- a/flake8_trio/visitors/flake8triovisitor.py
+++ b/flake8_trio/visitors/flake8triovisitor.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import ast
 from abc import ABC
-from typing import TYPE_CHECKING, Any, Union
+from typing import TYPE_CHECKING, Any, ClassVar, Union
 
 import libcst as cst
 from libcst.metadata import PositionProvider
@@ -23,7 +23,9 @@ if TYPE_CHECKING:
 
 class Flake8TrioVisitor(ast.NodeVisitor, ABC):
     # abstract attribute by not providing a value
-    error_codes: dict[str, str]  # pyright: ignore[reportUninitializedInstanceVariable]
+    error_codes: ClassVar[
+        dict[str, str]
+    ]  # pyright: ignore[reportUninitializedInstanceVariable]
 
     def __init__(self, shared_state: SharedState):
         super().__init__()
--- a/flake8_trio/visitors/flake8triovisitor.py
+++ b/flake8_trio/visitors/flake8triovisitor.py
@@ -23,9 +23,7 @@ if TYPE_CHECKING:
 
 class Flake8TrioVisitor(ast.NodeVisitor, ABC):
     # abstract attribute by not providing a value
-    error_codes: ClassVar[
-        dict[str, str]
-    ]  # pyright: ignore[reportUninitializedInstanceVariable]
+    error_codes: ClassVar[dict[str, str]]
 
     def __init__(self, shared_state: SharedState):
         super().__init__()
@@ -160,7 +158,7 @@ class Flake8TrioVisitor(ast.NodeVisitor, ABC):
 
 class Flake8TrioVisitor_cst(cst.CSTTransformer, ABC):
     # abstract attribute by not providing a value
-    error_codes: dict[str, str]  # pyright: ignore[reportUninitializedInstanceVariable]
+    error_codes: dict[str, str]
     METADATA_DEPENDENCIES = (PositionProvider,)
 
     def __init__(self, shared_state: SharedState):
--- a/flake8_trio/visitors/helpers.py
+++ b/flake8_trio/visitors/helpers.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 
 import ast
 from fnmatch import fnmatch
-from typing import TYPE_CHECKING, NamedTuple, TypeVar
+from typing import TYPE_CHECKING, NamedTuple, TypeVar, Union
 
 import libcst as cst
 import libcst.matchers as m
@@ -29,7 +29,9 @@ if TYPE_CHECKING:
 
     T = TypeVar("T", bound=Flake8TrioVisitor)
     T_CST = TypeVar("T_CST", bound=Flake8TrioVisitor_cst)
-    T_EITHER = TypeVar("T_EITHER", bound=Flake8TrioVisitor | Flake8TrioVisitor_cst)
+    T_EITHER = TypeVar(
+        "T_EITHER", bound=Union[Flake8TrioVisitor, Flake8TrioVisitor_cst]
+    )
 
 
 def error_class(error_class: type[T]) -> type[T]:
--- a/tests/autofix_files/trio911.py
+++ b/tests/autofix_files/trio911.py
@@ -13,8 +13,11 @@ async def foo() -> Any:
     await foo()
 
 
-def bar(*args) -> Any:
-    ...
+def bar(*args) -> Any: ...
+
+
+# mypy now treats `if ...` as `if True`, so we have another arbitrary function instead
+def condition() -> Any: ...
 
 
 async def foo_yield_1():
@@ -94,7 +97,7 @@ async def foo_async_for():  # error: 0, "exit", Statement("yield", lineno+4)
 async def foo_async_for_2():  # error: 0, "exit", Statement("yield", lineno+2)
     async for i in trio.trick_pyright:
         yield
-        if ...:
+        if condition():
             break
     await trio.lowlevel.checkpoint()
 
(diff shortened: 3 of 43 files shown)


jakkdl:
@thejcannon @aneeshusa

flake8_trio/visitors/helpers.py
```diff
@@ -29,7 +29,9 @@
 
     T = TypeVar("T", bound=Flake8TrioVisitor)
     T_CST = TypeVar("T_CST", bound=Flake8TrioVisitor_cst)
-    T_EITHER = TypeVar("T_EITHER", bound=Flake8TrioVisitor | Flake8TrioVisitor_cst)
+    T_EITHER = TypeVar(
+        "T_EITHER", bound=Union[Flake8TrioVisitor, Flake8TrioVisitor_cst]
+    )
```
thejcannon, line 34: Why this change?

tests/autofix_files/trio911.py
```diff
@@ -13,8 +13,11 @@ async def foo() -> Any:
     await foo()
 
 
-def bar(*args) -> Any:
-    ...
+def bar(*args) -> Any: ...
+
+
+# mypy now treats `if ...` as `if True`, so we have another arbitrary function instead
```
thejcannon, line 19: I mean...

```pycon
>>> bool(...)
True
```

😂

tests/autofix_files/trio91x_autofix.py
```diff
@@ -39,10 +38,10 @@ async def foo_yield():  # TRIO911: 0, "exit", Statement("yield", lineno+2)
 
 
 async def foo_if():
-    if ...:
+    if foo():
```
thejcannon, line 41: since `foo` is async, should we `if bar()` in these places?

tests/eval_files/trio102.py
```diff
@@ -120,7 +120,9 @@ async def foo():
     try:
         pass
     finally:
-        async for i in trio.bypasslinters:  # error: 8, Statement("try/finally", lineno-3)
+        async for (  # error: 8, Statement("try/finally", lineno-3)
+            i
+        ) in trio.bypasslinters:
```
thejcannon, line 125: What spurred this?

tests/eval_files/trio91x_autofix.py
```diff
@@ -34,9 +33,9 @@ async def foo_yield():  # TRIO911: 0, "exit", Statement("yield", lineno+2)
 
 
 async def foo_if():
-    if ...:
+    if foo():
```
thejcannon, line 36: Same comment about `async` `foo`

thejcannon (approved):
Mostly little things

Zac-HD (approved):
Thanks John!

flake8_trio/visitors/helpers.py
jakkdl: I didn't previously specify `python_version` in `pyproject.toml` for mypy, and now that I did it didn't like the old version which wouldn't work at runtime - and mypy doesn't seem to detect that we're inside an `if TYPE_CHECKING` block.

flake8_trio/visitors/flake8triovisitor.py
```diff
@@ -23,7 +23,9 @@
 
 class Flake8TrioVisitor(ast.NodeVisitor, ABC):
     # abstract attribute by not providing a value
-    error_codes: dict[str, str]  # pyright: ignore[reportUninitializedInstanceVariable]
+    error_codes: ClassVar[
+        dict[str, str]
+    ]  # pyright: ignore[reportUninitializedInstanceVariable]
```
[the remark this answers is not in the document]
jakkdl: It is not! And turns out `reportUnnecessaryTypeIgnoreComment` doesn't default to true with `strict = true`, so time to turn that on.
I did change this one hoping it would flow through and silence RUF012 ("Mutable class attribute should be annotated with `typing.ClassVar`") to the classes inheriting from it - but when that didn't happen I silenced the error rather than change it in the 25 other places. But maybe I'll bother doing the big search&replace

tests/autofix_files/trio91x_autofix.py
jakkdl: huh, I'd expect that to trigger [mypy's unused-coroutine](https://mypy.readthedocs.io/en/latest/error_code_list.html#check-that-coroutine-return-value-is-used-unused-coroutine) but I guess the return value is technically used.
Reported it at https://github.com/python/mypy/issues/16921#issue-2138489669

For the test file it of course doesn't really matter, unless mypy starts treating `if foo():` as `if True:` :grin:, but yeah no reason not to.

tests/eval_files/trio102.py
jakkdl: New version of Black now considers the line too long (90>88 after all) and splits it up, after which I move the comment back to line 123.
Not super pretty, maybe worth reporting upstream.

flake8_trio/visitors/flake8triovisitor.py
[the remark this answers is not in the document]
jakkdl: oh hm, the reason it wasn't enabled is because there's no separate setting for `pyright: ignore` vs `type: ignore` ([and unlikely to be added](https://github.com/python/mypy/issues/12358)), and [mypy doesn't have `mypy: ignore`](https://github.com/python/mypy/issues/12358)

flake8_trio/visitors/helpers.py
jakkdl: although I never remember which tools want `python_version` to be set to max-version-you-support vs min-version-you-support, so maybe should set it to 3.12 instead of 3.9

jakkdl: mypy wants you to run separate runs of all of them in parallel x)
https://github.com/python/mypy/issues/12286

flake8_trio/visitors/flake8triovisitor.py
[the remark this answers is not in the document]
jakkdl: Could perhaps enable mypy's `type: ignore`, but that'd involve adding a bunch of deps to its pre-commit env and/or go through a bunch of them and replace with `pyright: ignore` - and I don't think that's worth bothering with currently.

tests/eval_files/trio91x_autofix.py
jakkdl: yeah these are the same file, `eval_files` are the original files, and then the ones in in `autofix_files` are generated (and diff'd against in tests), so these two have to match.

flake8_trio/visitors/flake8triovisitor.py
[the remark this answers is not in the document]
Zac-HD: If we annotate as `Mapping[str, str]` so that it's logically immutable maybe our typecheckers will be happier?

tox.ini
```diff
@@ -1,23 +1,23 @@
...
 
 # create a default testenv, whose behaviour will depend on the name it's called with.
 # for CI you can call with `-e flake8_5,flake8_6` and let the CI handle python version
 [testenv]
 description = Runs pytest, optionally with posargs
 deps =
-    flake8_6: flake8>=6.0
-    flake8_5: flake8>=5.0,<6.0
+    flake8_7: flake8>=7.0
+    flake8_6: flake8>=6.0, <7.0
```
Zac-HD, line 12: Unclear from this whether or not we have flake8_5 and _7 environments - the deps list has changed, but not the envlist (or comment).  

Once you confirm that this is working as intended, including in CI, I'll merge 🙂

PR #203, opened 2024-02-21
Followup changes to fix ruff & pyright warnings

jakkdl:
* test_flake8_trio now subclasses `Statement` to add `__eq__`, so it's properly hashable elsewhere.
* `error_codes` is now annotated as `Mapping`, silencing RUF012
  * As suggested by @Zac-HD  in https://github.com/python-trio/flake8-trio/pull/202#discussion_r1492897532

Barely needs much of a review. @thejcannon @aneeshusa

### DIFF
--- a/tests/test_flake8_trio.py
+++ b/tests/test_flake8_trio.py
@@ -50,22 +50,6 @@ assert (
 class ParseError(Exception): ...
 
 
-class ColumnAgnosticStatement(Statement):
-    def __eq__(self, other: object) -> bool:
-        assert isinstance(other, Statement)
-        return (
-            self.name == other.name
-            and self.lineno == other.lineno
-            and (
-                self.col_offset == other.col_offset
-                or -1 in (self.col_offset, other.col_offset)
-            )
-        )
-
-    def __hash__(self) -> int:
-        raise NotImplementedError
-
-
 # check for presence of _pyXX, skip if version is later, and prune parameter
 def check_version(test: str):
     python_version = re.search(r"(?<=_PY)\d*", test)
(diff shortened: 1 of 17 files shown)


tests/test_flake8_trio.py
```diff
@@ -50,6 +50,22 @@
...
+            self.name == other.name
+            and self.lineno == other.lineno
+            and (
+                self.col_offset == other.col_offset
+                or -1 in (self.col_offset, other.col_offset)
+            )
+        )
+
+    def __hash__(self) -> int:
+        raise NotImplementedError
```
thejcannon, line 66: Why not `super().__hash__()`?

jakkdl: That could cause objects that are equal to have different hashes, which is Not Allowed :tm: 
https://docs.python.org/3/reference/datamodel.html#object.__hash__

Rereading it now, it turns out I misread that originally and I didn't have to complicate things as much as I did. I originally read it as `__eq__` objects had to have the same hash, and objects with same hash value had to be equal to eachother...

So I could implement `__hash__()` with `return hash(tuple(self.name, self.lineno))`

tests/test_flake8_trio.py
```diff
@@ -50,6 +50,22 @@
 class ParseError(Exception): ...
 
 
+class ColumnAgnosticStatement(Statement):
+    def __eq__(self, other: object) -> bool:
```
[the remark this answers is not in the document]
jakkdl: ended up reverting stuff and implementing a simple `__hash__` :)

Zac-HD (approved):

PR #204, opened 2024-02-21
use pre-commit.ci, remove workflow step running pre-commit. Add pyright to CI workflow

jakkdl:
Some minor shuffling around of things now that we have pre-commit.ci
A bit ugly to duplicate the typing dependencies, but since we don't pin those it's nbd.

pyright is in pre-commit primarily because I run it locally, but the pre-commit bot will now also bump versions of it, causing it to actually run in the CI action, and everything magically works out.

todo:
* edit the branch protection rules. Will do that just before merging.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,20 +8,23 @@ on:
         branches: [main]
 
 jobs:
-    check:
+    # pyright can't run in pre-commit CI as it requires an internet connection
+    # so we instead run it directly as a GitHub action
+    pyright:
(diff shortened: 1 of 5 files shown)


Zac-HD (approved):
LGTM, but since transferring the repo I don't have permissions to edit the branch protection rules 😅

jakkdl:
I do :grin:

jakkdl:
Removed `check` requirement, added 3.12, pyright, and pre-commit.

PR #205, opened 2024-02-23
remove trio117, as MultiError is removed in trio 0.24.0

jakkdl:
Good job past me, very easy to remove error codes :]


Zac-HD (approved):
Nice!

PR #206, opened 2024-02-23
Rename to flake8 async

jakkdl:
First steps in renaming the project:
1. Update documentation to refer to flake8-async instead of flake8-trio
2. Add name history to readme, expand it a bit in general.
3. rename the executable

In the second commit I rename all the error codes.
This was largely done with a search&replace of TRIO->ASYNC across all files, and then cleaning up the mess.
* several places assumed the length of the error code to be 7, which is now incorrect. (This is to turn e.g. TRIO103_anyio/TRIO103_trio into the actual error code. The `_anyio`/`_trio` is to generate different suggestion depending on which library you're using.)
  * The number 8 is now saved in a constant, or we use `re.match`, to avoid the magic number.

Third commit renames Flake8TrioVisitor to Flake8AsyncVisitor

Review the commits separately to maybe retain some sanity. I think the tests should be thorough enough to pick up anything problematic though, so don't think we have to review this super thoroughly.

Remaining steps:
* rename the repository to python-trio/flake8-async
  * update URL's. (I updated some to point to python-trio/flake8-trio instead of ZacHD/flake8-trio)
* Rename files/directories. I'll do this in a separate commit, maybe even a separate PR, once the content of files has been updated, since github fails to diff the content of a file if it's renamed in the same commit
  * Rename the eval & autofix files, removing the temporary renaming functions.
  * rename the `flake8_trio` directory, the `flake8_trio/visitors/flake8triovisitor.py` file, and the `tests/test_flake8_trio.py` file.


setup.py
```diff
@@ -51,7 +51,10 @@ def local_file(name: str) -> Path:
     ),
     long_description_content_type="text/markdown",
     entry_points={
-        "flake8.extension": ["TRI = flake8_trio:Plugin"],
-        "console_scripts": ["flake8-trio=flake8_trio:main"],
+        # You're not allowed to register error codes longer than 3 characters. But flake8
+        # doesn't enforce anything about the characters trailing the code, so we can say
+        # the code is ASY and then just always happen to print NCxxx directly after it.
```
Zac-HD, line 56: 😹

tests/autofix_files/noqa.py.diff
```diff
@@ -4,49 +4,49 @@
...
--    with trio.fail_after(5):  # TRIO100: 9, 'trio', 'fail_after'
--        yield  # TRIO911: 8, "yield", Statement("function definition", lineno-2)
+-    with trio.fail_after(5):  # ASYNC100: 9, 'trio', 'fail_after'
+-        yield  # ASYNC911: 8, "yield", Statement("function definition", lineno-2)
 +    await trio.lowlevel.checkpoint()
-+    # TRIO100: 9, 'trio', 'fail_after'
-+    yield  # TRIO911: 8, "yield", Statement("function definition", lineno-2)
++    # ASYNC100: 9, 'trio', 'fail_after'
++    yield  # ASYNC911: 8, "yield", Statement("function definition", lineno-2)
      await trio.lowlevel.checkpoint()
```
Zac-HD, line 12: Unintended semantic changes in this file?  And a few others below.

tests/test_messages_documented.py
```diff
@@ -14,8 +14,14 @@
 README = CHANGELOG.parent / "README.md"
 
 # 107 & 108 are removed (but still mentioned in changelog & readme)
-# TRIOxxx_* are fake codes to get different error messages for the same code
-IGNORED_CODES_REGEX = r"TRIO107|TRIO108|TRIO\d\d\d_.*"
+# ASYNCxxx_* are fake codes to get different error messages for the same code
+IGNORED_CODES_REGEX = r"(TRIO|ASYNC)107|(TRIO|ASYNC)108|ASYNC\d\d\d_.*"
```
Zac-HD, line 18:
```suggestion
IGNORED_CODES_REGEX = r"(TRIO|ASYNC)10[78]|ASYNC\d\d\d_.+"
```
I think this is net easier to read now

Zac-HD (commented):
Probably want to rebase on main to get the TRIO117 removal.  Overall looks great, and future rounds of review should be quicker with the "Viewed" status set for most files...

flake8_trio/visitors/flake8triovisitor.py
```diff
@@ -20,8 +20,11 @@
         ast.expr, ast.stmt, ast.arg, ast.excepthandler, ast.alias, Statement
     ]
 
+# 8 == len('ASYNCxxx'), so alt messages raise the original code
+ERROR_CODE_LEN = 8
```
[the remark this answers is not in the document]
jakkdl: oh yeah that's much better. I can split it out into a helper function too to standardize the functionality

setup.py
jakkdl: (this has always been the case, but noticed now it's probably good to add a comment)

tests/autofix_files/noqa.py.diff
jakkdl: Oh, *very* good catch. Need to investigate!

README.md
```diff
@@ -1,122 +1,124 @@
...
 
 ```console
-pip install flake8-trio
+pip install flake8-async
 ```
 
 ## List of warnings
 
-- **TRIO100**: A `with trio.fail_after(...):` or `with trio.move_on_after(...):`
+- **ASYNC100**: A `with trio.fail_after(...):` or `with trio.move_on_after(...):`
```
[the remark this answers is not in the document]
jakkdl: Good point yeah, I'll do a comprehensive review which checks work with which library when I go through them for asyncio. #208

tests/autofix_files/noqa.py.diff
jakkdl: Reminder what the different files are for, from `CONTRIBUTING.md`:
> ### `# AUTOFIX`
> Files in `tests/eval_files` with this marker will have two files in `tests/autofix_files/`. One with the same name containing the code after being autofixed, and a diff file between those two.  
> During tests the result of running the checker on the eval file with autofix enabled will be compared to the content of the autofix file and will print a diff (if `-s` is on) and assert that the content is the same. `--generate-autofix` is added as a pytest flag to ease development, which will print a diff (with `-s`) and overwrite the content of the autofix file.  
> Files without this marker will be checked that they *don't* modify the file content.


Theree are no unintended changes in `tests/eval_files/noqa.py`, nor in `tests/autofix_files/noqa.py`, so there *shouldn't* be unintended changes in `tests/autofix_files/noqa.py.diff`. And when actually looking at https://github.com/python-trio/flake8-trio/blob/022fc3a4f661ec42afac5a21eb0404fe13f62380/tests/autofix_files/noqa.py.diff and comparing it to https://github.com/python-trio/flake8-trio/blob/780e5cd685e1f4ee187064567aca5890f4a7c0c4/tests/autofix_files/noqa.py.diff there in fact aren't any unintended changes.

Looking closer at the diff, the confusion stems from `.diff` files being syntax-highlighted, so we're viewing a diff of a `.diff` file. The solid green/red background, and the righmost +/-, is the diff from the file, and the line number having a green/red background & the leftmost +/- is the content of the diff file actually being changed in this commit. So e.g. line 9 is *not* changed in the commit, and there are no semantic changes.

The green on line 9 is to indicate that when running autofix on `eval_files/noqa.py` we have a checkpoint added.

tests/test_messages_documented.py
```diff
@@ -14,8 +14,14 @@
 README = CHANGELOG.parent / "README.md"
 
 # 107, 108 & 117 are removed (but still mentioned in changelog & readme)
-# TRIOxxx_* are fake codes to get different error messages for the same code
-IGNORED_CODES_REGEX = r"TRIO107|TRIO108|TRIO117|TRIO\d\d\d_.*"
+# ASYNCxxx_* are fake codes to get different error messages for the same code
+IGNORED_CODES_REGEX = r"(TRIO|ASYNC)(107|108|117)|ASYNC\d\d\d_.*"
```
Zac-HD, line 18: Much nicer than my proposal 👍

tests/test_messages_documented.py
```diff
@@ -26,39 +32,44 @@ def test_messages_documented():
         filename = path.name
         documented_errors[filename] = set()
         for line in lines:
-            for error_msg in re.findall(r"TRIO\d\d\d", line):
-                documented_errors[filename].add(error_msg)
+            for error_msg in re.findall(r"TRIO\d\d\d|ASYNC\d\d\d", line):
+                print(error_msg, rename_trio_to_async(error_msg))
```
Zac-HD, line 36: Stray `print()`?

Zac-HD (approved):
LGTM with one nit about a `print()`, merge when ready 🎉 

And after that I'll update the PyPI token, bump version, and release under `flake8-async`!  (and once we've seen everything working smoothly for a while, I'll make a final placeholder/depender release under `flake8-trio`)

tests/test_messages_documented.py
jakkdl: good catch, removed.

PR #209, opened 2024-02-26
rename files, update references to them

jakkdl:
Some straightforward search&replace after #206


jakkdl:
All eval files are named `trioXXX.py`, so they should maybe all be renamed as well. Although with the given design they are written with `trio` as the baseline, running tests against `anyio` by doing search/replace on the file content. So there's a decent case for the status quo, although there are files named `trio104_trio.py` or `trio105_anyio.py`...

Zac-HD (approved):

Issue #208, opened 2024-02-25
Add asyncio support to current checks


jakkdl:
* Go through all errors and see which ones are applicable to asyncio
  * Add support & parametrize tests for them
* Update README.md to indicate which tests are available for which library (asyncio, anyio, trio)

Could also consider supporting curio or any other IO library.

jakkdl:
I'm not familiar with asyncio, but here's a first pass through all our errors on which ones I think are applicable to asyncio:

### no
- **ASYNC100**: A `with trio.fail_after(...):` or `with trio.move_on_after(...):`
  context does not contain any `await` statements.  This makes it pointless, as
  the timeout can only be triggered by a checkpoint.
- **ASYNC101**: `yield` inside a nursery or cancel scope is only safe when implementing a context manager - otherwise, it breaks exception handling.
- **ASYNC105**: Calling a trio async function without immediately `await`ing it.
  - now that we know type checkers handle this I'm unenthused about expanding this check.
- **ASYNC109**: Async function definition with a `timeout` parameter - use `trio.[fail/move_on]_[after/at]` instead
- **ASYNC111**: Variable, from context manager opened inside nursery, passed to `start[_soon]` might be invalidly accessed while in use, due to context manager closing before the nursery. This is usually a bug, and nurseries should generally be the inner-most context manager.
- **ASYNC112**: Nursery body with only a call to `nursery.start[_soon]` and not passing itself as a parameter can be replaced with a regular function call.
- **ASYNC113**: Using `nursery.start_soon` in `__aenter__` doesn't wait for the task to begin. Consider replacing with `nursery.start`.
- **ASYNC114**: Startable function (i.e. has a `task_status` keyword parameter) not in `--startable-in-context-manager` parameter list, please add it so ASYNC113 can catch errors when using it.
- **ASYNC115**: Replace `trio.sleep(0)` with the more suggestive `trio.lowlevel.checkpoint()`.
- **ASYNC116**: `trio.sleep()` with >24 hour interval should usually be `trio.sleep_forever()`.
- **ASYNC118**: Don't assign the value of `anyio.get_cancelled_exc_class()` to a variable, since that breaks linter checks and multi-backend programs.


### maybe
- **ASYNC102**: It's unsafe to await inside `finally:` or `except BaseException/trio.Cancelled` unless you use a shielded cancel scope with a timeout.
  - I don't know if there's any "shielded cancel scopes with a timeout", but awaiting inside finally/baseexception might be bad?
[...]

PR #210, opened 2024-02-26
basic Asyncio support

jakkdl:
first steps on #208 
106, 200, 21x and 9xx are supported, though no autofix.

This is rebased on top of #209, so if reviewing you should only look at the latest commit(s).

todo:
* [x] update readme.md
* [x] update changelog
* [x] do a second pass through the diff, quite tired atm so suspect there may be silly stuff somewhere.
* [ ] rewrite descriptions of --anyio / --asyncio to not be dumb. And/or have a `--library` flag instead.


Zac-HD:
Looks good so far!  Will re-review when you think it's done 👍

tests/test_flake8_async.py
```diff
@@ -195,10 +188,16 @@ def check_autofix(
...
 @dataclass
 class MagicMarkers:
+    # Exclude checking a library against a file
     NOANYIO: bool = False
     NOTRIO: bool = False
+    NOASYNCIO: bool = False
+    # File should raise no errors with this library
     ANYIO_NO_ERROR: bool = False
-    ASYNC_NO_ERROR: bool = False
+    TRIO_NO_ERROR: bool = False
```
jakkdl, line 197: this is a revert of an overeager search+replace in previous PR.

flake8_async/visitors/visitor_utility.py
```diff
@@ -117,11 +117,13 @@ def __init__(self, *args: Any, **kwargs: Any):
         # see imports
         if self.options.anyio:
             self.add_library("anyio")
+        if self.options.asyncio:
+            self.add_library("anyio")
```
Zac-HD, line 121:
```suggestion
        if self.options.asyncio:
            self.add_library("asyncio")
```

Zac-HD (commented):
let's rebase on master, add `ASYNC102` support, and then we're probably done?

flake8_async/visitors/visitor_utility.py
jakkdl: huh, surprised this was not caught by a linter for "different branches of if with same content should be combined with `or`"

jakkdl: oh derp, it's because they're separate `if`s - not different branches for the same `if`.

jakkdl:
Uh, I ended up doing A Lot:tm:. But the diff didn't actually end up being all that big. The rebase might break viewing only new changes though, so you might want to review commit-by-commit.

Only ones listed in #208 not supported now is:
* 22X/23X/24X not having tailored error messages (and therefore being untested)
* not detecting asyncio-shielded-timeouts

So I'm now comfortable saying this has decent `asyncio` support and we could push a release.

TODO: Oh I did rename the repository, so should update links.

tests/test_config_and_args.py
```diff
@@ -174,13 +174,24 @@ def test_anyio_from_config(tmp_path: Path, capsys: pytest.CaptureFixture[str]):
...
     )
-    err_file = str(Path(__file__).parent / "eval_files" / "anyio_trio.py")
-    expected = f"{err_file}:10:5: ASYNC220 {err_msg}\n"
+    err_file = Path(__file__).parent / "eval_files" / "anyio_trio.py"
+
+    # find the line with the expected error
+    for i, line in enumerate(err_file.read_text().split("\n")):
+        if "# ASYNC220: " in line:
+            # line numbers start at 1, enumerate starts at 0
+            lineno = i + 1
```
Zac-HD, line 183:
```suggestion
    for lineno, line in enumerate(err_file.read_text().split("\n"), start=1):
        if "# ASYNC220: " in line:
```

https://docs.python.org/3/library/functions.html#enumerate

Zac-HD (approved):
LGTM; agree that we could release this and should release whatever we have on the weekend (when I set up release automation again).

I'll let you merge after any docs/links updates you want to make.

tests/test_config_and_args.py
jakkdl: :heart_eyes:

jakkdl:
ah oops, I realized I forgot to ask your opinion on `use [trio|anyio].lowlevel.checkpoint()` vs `use [trio/anyio].lowlevel.checkpoint()` (or something else). Both are used in different places in documentation/comments/error messages

Zac-HD:
I'd probably go with the slash, but don't really care so long as we're consistent.

PR #216, opened 2024-03-08
add ASYNC250: blocking sync call input() in async function.

### LINKED ISSUE #199: New rule: builtin `input()` should not be called from async function
```python
async def fn():
    ans = input("what to say")  # bad - no task can run while waiting for the user
    ans = await trio.to_thread.run_sync(input, "what to say")  # good, if inelegant
```

We could add this to TRIO222, or give it a new error code specific to this function.

jakkdl:
fixes #199


Zac-HD (approved):
🎉

PR #219, opened 2024-03-15
Add new check ASYNC251 for time.sleep()

jakkdl:
As noted in https://github.com/astral-sh/ruff/pull/10416 old-flake8-async checked for `time.sleep()`, but none do currently, so I quickly added it.


Zac-HD (approved):
Thanks for the quick fix!

Issue #156, opened 2023-03-14
Should `ASYNC102` special-case `.aclose()`?


Zac-HD:
[`trio.abc.AsyncResource.aclose()` specifies](https://trio.readthedocs.io/en/stable/reference-io.html#trio.abc.AsyncResource) that it must close the resource *even if the method is cancelled or raises an exception*; the error indicates a failure to close gracefully but the resource will still be closed.

This means that it *is* safe to write `finally: await x.aclose()` - in the no-error case it will close gracefully; and if cancelled it's equivalent to `trio.aclose_forcefully(x)` and will still close.  Adding a cancel-shielded scope and timeout here is actually counter-productive!

So... should we special-case this method?  Or perhaps rethink the check more generally?  I'm not sure, but wanted to track the question.  It's not an abstract question either, if we want to lint for e.g. https://github.com/encode/httpcore/issues/642#issuecomment-1459636877.

jakkdl:
wow this is some complicated shit, but I've hopefully succeeded in wrapping my head around it sufficiently to have a basic grasp.
Excluding it from trio102 sounds correct at least, but it may definitely need more logic and/or another check to handle the various cases.
Shouldn't be hard to invert the check specifically for `finally: await x.aclose()` and give a warning if that's inside a [cancel]-[shielded] scope.

Other than aclose or complicated teardown stuff, is the base-case trio102 useful?

But yeah whatever you conclude I can almost certainly implement it, the variable- and typetracking (+me having dozens of visitors under my belt) should be powerful enough for even very weird cases, and this sounds important enough that a lot of logic is actually gonna be worth it.

Zac-HD:
Fixed by #222

PR #222, opened 2024-03-19
ASYNC102 no longer raises warning for calls to `.aclose()` on objects

jakkdl:
See #156 

I'm not sure if #156 also covers more general discussion on if/how to overhaul async102 and should remain open after this is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 *[CalVer, YY.month.patch](https://calver.org/)*
 
+## 24.3.4
+- ASYNC102 (no await inside finally or critical except) no longer raises warnings for calls to `aclose()` on objects. See https://github.com/python-trio/flake8-async/issues/156
+
 ## 24.3.3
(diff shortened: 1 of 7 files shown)


Zac-HD:
Very quick notes:
- we should allow this in an `except` clause too
- only exempt if the call takes no arguments, i.e. still error on `await x.aclose(foo)`
- merge conflict from the other PR 😅

jakkdl:
> don't exempt asyncio-only code; this is a trio/anyio semantics thing

I interpreted this as "if running with --anyio and not importing anyio/trio, or only importing asyncio => don't mark as safe; but if running with --anyio and/or importing anyio/trio *and* importing asyncio => mark as safe", so if you have a mixed project it'll lean towards not reporting `aclose`s. But I could make it more aggressive and warn if there's any risk that this part concerns asyncio-code (i.e. `asyncio in self.library` -> "running with --asyncio; or asyncio is imported")

Zac-HD (approved):
Nice!  Thanks, @jakkdl 🙂

Issue #214, opened 2024-03-05
Generate proper documentation


jakkdl:
README.md has grown to 200 lines, and spans error codes, installation & run instructions, configuration and command-line flags. We also have CHANGELOG.md with 150 lines.
There are a few links in the readme, but it would probably be quite useful to have anything referring to [objects within] urllib3, httpx, anyio, asyncio, trio, etc be hyperlinks w/o having to manually update.
Pulling out the error messages for the documentation automatically would be nice, but somewhat complicated given the current setup for messages differing for different (combinations of) libraries. Should be quite doable though.

https://github.com/python-trio/flake8-async/issues/211#issuecomment-1971778351
> I'm planning to write up some more detail about each of our error codes in a long doc, hopefully this weekend, which will include much more about why you'd want to enable each of the optional rules in addition to the existing material.

This also sounds like great doc material.

The one workflow I'm decently accustomed to is ReadTheDocs+Sphinx as used in python-trio/trio, and probably nice to be consistent with it since it's in the same organization. But it's also been a huge pain to work with at times.

Zac-HD:
I think an RTD/Sphinx setup with default config could be good.  One intro page with usage instructions (inc CLI); one page with all the error codes including explanatory notes; one page for contributor docs etc.

(still hoping for some free time on _a_ weekend... we'll see)

jakkdl:
After #245 the only thing remaining is expanding the rules to subpages with examples/explanatory notes/etc.

jakkdl:
Remaining tasks after #248 (as it currently is):
* [ ] Write separate subpages for rules
* [ ] adjust page width for `rules.rst`
* extensions:
  * [ ] hoverxref
  * [x] codeautolink
  * [x] intersphinx

PR #223, opened 2024-03-20
add basic sphinx & readthedocs infra

jakkdl:
Let's see if this works~


jakkdl:
next up is installing requirements, or scraping the `__version__` line manually if we're not gonna need to access the package for other reasons.
And then copy over a bunch of stuff from the readme and actually write/LLM-generate some docs

jakkdl:
Nice. Now I just gotta get some docs

jakkdl:
Slowly closing in on this being an improvement/on-par to status quo.

jakkdl:
Okay I think this is worth reviewing now. You can view the docs at https://flake8-async--223.org.readthedocs.build/en/223/ (or click through from the CI RTD check)
Next step would be breaking out rules-specific pages with a la ruff, but I'd need to spend some solid time figuring out a decent way of doing that which doesn't involve me manually creating a million different `.rst` files. I've been digging through the source for the rust docs to try and see how they do it but not been able to so far.

Other minor TODOs:
* Move `CONTRIBUTING.md` to the docs
* Get the docs linked in the readme, on PyPI, etc
* Strip down `README.md`
* Push a release

But I think we can put those in a separate PR. Would be nice to get this merged so I can update docs/ when writing feature PRs.

docs/usage.rst
```diff
@@ -0,0 +1,207 @@
...
+^^^^^^^^
+
+.. code-block:: zsh
+
+   flake8-async **/*.py
+
+
+Run through ruff
+================
+`Ruff <https://github.com/astral-sh/ruff>` is a linter and formatter that reimplements a lot of rules from various flake8 plugins. They currently only support a small subset of the rules though, see https://github.com/astral-sh/ruff/issues/8451 for current status and https://docs.astral.sh/ruff/rules/#flake8-async-async for documentation.
```
Zac-HD, line 80:
```suggestion
`Ruff <https://github.com/astral-sh/ruff>`_ is a linter and formatter that reimplements a lot of rules from various flake8 plugins.

They currently only support a small subset of the rules though, see https://github.com/astral-sh/ruff/issues/8451 for current status and https://docs.astral.sh/ruff/rules/#flake8-async-async for documentation.
```

docs/usage.rst
```diff
@@ -0,0 +1,207 @@
...
+
+*************
+Configuration
+*************
+
+`You can configure flake8 with command-line options <https://flake8.pycqa.org/en/latest/user/configuration.html>`_,
+but we prefer using a config file. The file needs to start with a section marker ``[flake8]`` and the following options are then parsed using flake8's config parser, and can be used just like any other flake8 options.
```
Zac-HD, line 87: Specify allowed filenames / types?

Zac-HD (approved):
See comments above, happy for you to merge this when you think it's ready and follow up later.

docs/rules.rst
```diff
@@ -0,0 +1,55 @@
...
+- **ASYNC911**: Exit, ``yield`` or ``return`` from async iterable with no guaranteed checkpoint since possible function entry (yield or function definition)
+  Checkpoints are ``await``, ``async for``, and ``async with`` (on one of enter/exit).
+
+Removed warnings
+~~~~~~~~~~~~~~~~
+
+- **TRIOxxx**: All error codes are now renamed ASYNCxxx
+- **TRIO107**: Renamed to TRIO910
+- **TRIO108**: Renamed to TRIO911
+- **TRIO117**: Don't raise
```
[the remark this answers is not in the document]
jakkdl: smh, that's Claude randomly deciding to stop giving output without any warning :upside_down_face:

docs/rules.rst
```diff
@@ -0,0 +1,55 @@
+List of warnings
+------------------
+
+- **ASYNC100**: A ``with [trio/anyio].fail_after(...):`` or ``with [trio/anyio].move_on_after(...):`` context does not contain any ``await`` statements.  This makes it pointless, as the timeout can only be triggered by a checkpoint. This check also allows ``yield`` statements, since checkpoints can happen in the caller we yield to.
```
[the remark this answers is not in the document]
jakkdl: > Let's add a "General rules" subheading for the 1xx rules.

:+1:

> I'd also consider using a sub-subheading for each individual rule (instead of the bullet points), so that it's easier to add code examples etc. That's probably better as a follow-up PR though.

It's either that or a separate subpage for each one (ruff style) as you suggested in https://github.com/python-trio/flake8-async/issues/214#issuecomment-1981529208

I'm not sure which one I prefer, tons of small pages can be messy to navigate - so maybe a dense list for an overview that then links to longer sections for each rule. A raw TOC wouldn't be great with error codes, but we probably should come up with human-friendly names for the codes at some point (https://github.com/astral-sh/ruff/issues/1773)

But yeah, leaving that one for followup as I said in https://github.com/python-trio/flake8-async/pull/223#issuecomment-2056378770

docs/usage.rst
jakkdl: rewrote/expanded it, referring to flake8 documentation for filenames/types/section marker. Also linking pyproject-flake8, and suggesting various ways of getting around us not having config support for running as a standalone.

PR #228, opened 2024-04-01
ASYNC100 no longer triggers on context managers containing a `yield`

### LINKED ISSUE #224: `ASYNC100` should not trigger for blocks containing a `yield` statement
```python
@contextmanager
def fail_at_with_error(deadline, error):
    with trio.move_on_at(deadline) as cancel_scope:  # noqa: ASYNC100
        yield cancel_scope
    if cancel_scope.cancelled_caught:
        raise error
```

As of the current version, the `noqa` comment is suppressing an error.  However, the `move_on_at()` _could_ be triggered by checkpoints syntactically outside this block, so I think we should avoid false alarms and allow this case.

jakkdl:
I .. don't know why I didn't open a PR when pushing this branch. fixes #224

### DIFF
--- a/tests/autofix_files/async100.py
+++ b/tests/autofix_files/async100.py
@@ -114,3 +114,9 @@ async def more_nested_tests():
     with contextlib.suppress(Exception):
         with open("blah") as file:
             print("foo")
+
+
+# Don't trigger for blocks with a yield statement
+async def foo():
+    with trio.fail_after(1):
+        yield
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 *[CalVer, YY.month.patch](https://calver.org/)*
 
+## 24.3.6
+- ASYNC100 no longer triggers if a context manager contains a `yield`.
+
 ## 24.3.5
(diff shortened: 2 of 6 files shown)


jakkdl:
oh that's why, it wasn't finished x)

Zac-HD (approved):
Thanks!

tests/autofix_files/async100.py
```diff
@@ -114,3 +114,9 @@ async def more_nested_tests():
     with contextlib.suppress(Exception):
         with open("blah") as file:
             print("foo")
+
+
+# Don't trigger for blocks with a yield statement
+async def foo():
+    with trio.fail_after(1):
+        yield
```
Zac-HD, line 122: Ah.  This should indeed be allowed by `ASYNC100`...

...but now I'm wondering about a non-optional rule to warn when you yield across a known cancelscope and the function isn't a context-manager, [because that really doesn't work](https://discuss.python.org/t/preventing-yield-inside-certain-context-managers/1091).  Probably not worth the trouble, since njs and I are planning to get runtime enforcement, the rule would have many missed alarms, and `ASYNC900` covers it anyway for people who care.

Zac-HD (approved):

PR #232, opened 2024-04-09
fix internal error in visitor91x

### LINKED ISSUE #226: Internal error with `ASYNC91x`
```python
import trio

async def fn():
    try:
        while True:
            try:
                await trio.sleep(1)
            except Exception:
                pass
    except Exception:
        pass
```
triggers
```
  File ".../flake8_async/visitors/visitor91x.py", line 346, in check_function_exit
    assert not isinstance(original_node, cst.FunctionDef)
  AssertionError
```

jakkdl:
fixes #226 
Underlying issue was different try/except's sharing the same reference to self.try_state, due to not passing `copy=True` to `save_state()`.

I went through the other calls to `save_state` in the same visitor and added an explicit `copy=False` to the only other one that didn't already specify it. It's possible `save_state()` shouldn't have a default at all, or default to `True` since the performance loss from some extra copying would probably be negligible.

### DIFF
--- a/tests/autofix_files/async910.py.diff
+++ b/tests/autofix_files/async910.py.diff
@@ -205,8 +205,16 @@
 
 
  async def foo_comprehension_3():
-@@ x,3 x,4 @@
+@@ x,6 x,7 @@
 
  async def await_everywhere_except_gen_target():  # error: 0, "exit", Statement("function definition", lineno)
      (await x async for x in bar())
++    await trio.lowlevel.checkpoint()
+
+
+ # Issue #226
+@@ x,3 x,4 @@
+                 pass
+     except Exception:
+         pass
 +    await trio.lowlevel.checkpoint()
--- a/tests/test_flake8_async.py
+++ b/tests/test_flake8_async.py
@@ -416,6 +416,7 @@ def _parse_eval_file(
                 msg = f"Line {lineno}: Failed to format\n {message!r}\nwith\n{args}"
                 raise ParseError(msg) from e
 
+    assert enabled_codes, "no codes enabled. Fix file name or add `# ARGS --enable=...`"
     enabled_codes_list = enabled_codes.split(",")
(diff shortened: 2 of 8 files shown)


tests/autofix_files/async910.py.diff
```diff
@@ -205,8 +205,16 @@
 
 
  async def foo_comprehension_3():
-@@ x,3 x,4 @@
+@@ x,6 x,7 @@
```
jakkdl, line 208: Ooooh after some confusion I think I figured out why this line in the diff output got changed. It's because there's now context after the changed line in `await_everywhere_except_gen_target` (i.e. two empty lines and `# Issue #226`), so now there's 3 more lines in the diff output.

tests/autofix_files/async910.py.diff
```diff
@@ -205,8 +205,16 @@
...
      (await x async for x in bar())
 +    await trio.lowlevel.checkpoint()
+
+
+ # Issue #226
+@@ x,3 x,4 @@
+                 pass
+     except Exception:
+         pass
++    await trio.lowlevel.checkpoint()
```
jakkdl, line 220: once again this looks kinda weird in the github diff view, due to viewing a syntax-highlighted diff file in diff view. But the actual diff is entirely as expected.

tests/test_flake8_async.py
```diff
@@ -416,6 +416,7 @@ def _parse_eval_file(
                 msg = f"Line {lineno}: Failed to format\n {message!r}\nwith\n{args}"
                 raise ParseError(msg) from e
 
+    assert enabled_codes, "no codes enabled. Fix file name or add `# ARGS --enable=...`"
```
jakkdl, line 419: minor unrelated QoL when writing new tests to get a nice error message.

Zac-HD (approved):
Sweet, thanks @jakkdl!

Issue #211, opened 2024-02-27
Suggestion: `ASYNC119` when using an async context manager in an async generator


alicederyn:
**Antipattern:** The following code is unsafe:

```python
async def yield_some_values():
    async with some_async_context_manager():
        for _ in range(5):
            ...
            yield some_value  # error: `yield` inside `async with` block

async def bad_method():
    async for value in yield_some_values():
        raise Exception("exiting early")
```

**Explanation:** If the iterator is not fully consumed, the cleanup path will only be triggered by the garbage collector, which will not allow the async cleanup logic to `await` (it will raise `GeneratorExit` when it tries). A similar issue arises if you try to `await` in a `finally` block in an async generator, or an `except` block that can intercept `asyncio.exceptions.CancelledError`.

The only exception to this that I'm aware of is if the async generator is decorated with `@contextlib.asynccontextmanager`.

**Fix:** I think unfortunately the only fix is to refactor the API to separate the context management from the iteration, e.g.

```python
@contextlib.asynccontextmanager
async def yield_some_values_safely():
    async with some_async_context_manager():
        async def inner_iterator():
            for _ in range(5):
                ...
                yield some_value
        yield inner_iterator  # safe because this is in an asynccontextmanager

async def bad_method():
    async with yield_some_values_safely() as values:
        async for value in values:
            raise Exception("exiting early")
```

I'm not aware of any existing lint check for this antipattern (I may have missed it though!) &mdash; do the maintainers of this repo think it would be valuable?

alicederyn:
Actually, I'm wondering if this is exactly `ASYNC101` and I just don't understand what a "nursery or cancel scope" is? In which case this may just be a request to clarify the wording in the README to say "async with" block as well?

Zac-HD:
Oh boy, this is much harder than it sounds - see https://peps.python.org/pep-0533/ and https://github.com/python-trio/trio/issues/265 for some of the details.  *Additionally* you can get similar issues if you `yield` across a `CancelScope` (a _sync_ context manager), including having exceptions including cancellations delivered to the wrong task 😱 

My personal view is that async generators are simply too dangerous and too difficult to use correctly, and so at work we've banned them entirely via the `TRIO900` lint rule - soon to be `ASYNC900`.  To make async generators reasonably safe, I think we'd need to

[...]

PR #238, opened 2024-04-23
add async119: yield in contextmanager in async generator

jakkdl:
See #211 
@alicederyn

The issue is somewhat confusing at times to parse out what discussing pertains to async119 vs async102 vs hypotheticals, but I think this is what was settled upon.

I *think* this is a false positive with current implementation
```python
async def foo():
    with open(""):
      yield
```
i.e. the contextmanager is sync (and there are no awaits after the contextmanager?). Resolving that wouldn't be terribly complicated so I'll implement that if somebody confirms my understanding.

I didn't put much energy in formatting the entry in the readme, as that is on its way out. The links in the docs would probably be much cleaner with intersphinx or something, but leaving that for a different PR.

### DIFF
--- a/docs/rules.rst
+++ b/docs/rules.rst
@@ -45,10 +45,11 @@ Note: 22X, 23X and 24X has not had asyncio-specific suggestions written.
 - **ASYNC250**: Builtin ``input()`` should not be called from async function. Wrap in ``[trio/anyio].to_thread.run_sync()`` or ``asyncio.loop.run_in_executor()``.
 - **ASYNC251**: ``time.sleep(...)`` should not be called from async function. Use ``[trio/anyio/asyncio].sleep(...)``.
 
-.. _async900:
 
 Optional rules disabled by default
 ==================================
+
+.. _async900:
 
 - **ASYNC900**: Async generator without ``@asynccontextmanager`` not allowed. You might want to enable this on a codebase since async generators are inherently unsafe and cleanup logic might not be performed. See https://github.com/python-trio/flake8-async/issues/211 and https://discuss.python.org/t/using-exceptiongroup-at-anthropic-experience-report/20888/6 for discussion.
 - **ASYNC910**: Exit or ``return`` from async function with no guaranteed checkpoint or exception since function definition. You might want to enable this on a codebase to make it easier to reason about checkpoints, and make the logic of ASYNC911 correct.
--- a/docs/rules.rst
+++ b/docs/rules.rst
@@ -21,6 +21,9 @@ General rules
 - **ASYNC115**: Replace ``[trio/anyio].sleep(0)`` with the more suggestive ``[trio/anyio].lowlevel.checkpoint()``.
 - **ASYNC116**: ``[trio/anyio].sleep()`` with >24 hour interval should usually be ``[trio/anyio].sleep_forever()``.
 - **ASYNC118**: Don't assign the value of ``anyio.get_cancelled_exc_class()`` to a variable, since that breaks linter checks and multi-backend programs.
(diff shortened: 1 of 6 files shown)


Zac-HD:
> I _think_ this is a false positive with current implementation
> 
> ```python
> async def foo():
>     with open(""):
>       yield
> ```
> 
> i.e. the contextmanager is sync (and there are no awaits after the contextmanager?).

No, that's not a false alarm: 
```python
import contextlib, itertools, trio
resource_id = itertools.count()

@contextlib.contextmanager
def hold_some_resource():
    n = next(resource_id)
    print(f"acquire {n=}")
    try:
        yield
    finally:
        print(f"release {n=}")

async def loop_with_resource():
    with hold_some_resource():
        yield
        yield

@trio.run
async def main():
    for n in range(3):
        async for _ in loop_with_resource():
            if n == 0:
                break
```
```
acquire n=0
acquire n=1
release n=1
acquire n=2
release n=2
release n=0
```
and this can still happen with a single un-looped yield, depending on how it's called (ie if you create the generator but never iterate it, or iterate into the context but not out)

docs/rules.rst
```diff
@@ -42,6 +45,8 @@ Note: 22X, 23X and 24X has not had asyncio-specific suggestions written.
 - **ASYNC250**: Builtin ``input()`` should not be called from async function. Wrap in ``[trio/anyio].to_thread.run_sync()`` or ``asyncio.loop.run_in_executor()``.
 - **ASYNC251**: ``time.sleep(...)`` should not be called from async function. Use ``[trio/anyio/asyncio].sleep(...)``.
 
+.. _async900:
```
Zac-HD, line 48:
```suggestion
.. _async9xx:
```
to make it clear this is for the category

flake8_async/visitors/visitors.py
```diff
@@ -281,6 +281,47 @@ def visit_Call(self, node: ast.Call):
...
+        self.save_state(node, "contextmanager")
+        self.contextmanager = node
+
+    def visit_Yield(self, node: ast.Yield):
+        if self.unsafe_function is not None and self.contextmanager is not None:
+            # Decision point: the error could point to the method, or context manager,
+            # or the yield.
+            self.error(node)
+            # only warn once per method (?)
+            self.unsafe_function = None
```
Zac-HD, line 318: I think warning on each yield which is inside a context manager would be best; each one is a separate chance of problems.

Zac-HD (approved):
Minor comments, but looks great overall - merge when ready!

alicederyn:
> No, that's not a false alarm:

I'm confused, your example seems to show that it *is* a false alarm, and that's also my understanding.

docs/rules.rst
jakkdl: I wanted to link to 900 in particular but couldn't get the reference to work... but trying it again now it works without issue so I guess I'll just move it there :upside_down_face:

jakkdl:
> > No, that's not a false alarm:
> 
> I'm confused, your example seems to show that it _is_ a false alarm, and that's also my understanding.

The issue is that the release of 0 is delayed until the end, no? If we remove the break from inside the loop we get
```
acquire n=0
release n=0
acquire n=1
release n=1
acquire n=2
release n=2
```
which seems like sensible behaviour

alicederyn:
> The issue is that the release of 0 is delayed until the end

Cleanup being delayed until the destructor is called is potentially surprising, but I'm not sure I'd call it a bug? To me, the really big problem is when cleanup doesn't happen at all because the cleanup code cannot await in the destructor, and I was expecting that to be the case we focus on.

> I thought this could be reproduced by changing the generator to a sync one, but it seems to only happen with async generators in particular.

At a guess, I suspect this will be due to cleanup being done on refcount decrement for one but only in the garbage collector on the other &mdash; perhaps putting a `gc.collect()` call in will clarify.

Zac-HD:
Yeah, `__(a)exit__` is delayed until garbage collection, which can be arbitrarily late - including after the loop is shut down, and calling `gc.collect()` doesn't help in general because someone might still be holding a reference.

This is why https://peps.python.org/pep-0533/ exists, and why I'm speaking about it at the language summit next month!

Issue #215, opened 2024-03-05
[anyio/asyncio].TaskGroup support for 101, 111, 112


jakkdl:
These tests are currently hardcoded to check for nurseries, but should be extended to also check anyio taskgroups. The same errors are likely relevant for asyncio as well, so can extend to asyncio taskgroups too.

Errors in question are:
- **ASYNC101**: `yield` inside a trio/anyio nursery or cancel scope is only safe when implementing a context manager - otherwise, it breaks exception handling.
- **ASYNC111**: Variable, from context manager opened inside nursery, passed to `start[_soon]` might be invalidly accessed while in use, due to context manager closing before the nursery. This is usually a bug, and nurseries should generally be the inner-most context manager.
- **ASYNC112**: Nursery body with only a call to `nursery.start[_soon]` and not passing itself as a parameter can be replaced with a regular function call.

jakkdl:
It's possible there are other checks where I've missed the nursery/anyio distinction, I'll make a pass to check for any other that seem to only handle nurseries.

PR #241, opened 2024-05-01
add asyncio/anyio taskgroup support to async111&112, update contributing.md

jakkdl:
Fixes 2/3 tasks in #215

The logic in Visitor112 is incredibly messy and ugly, but works. Might rewrite it after sleeping on it.

I also realized that async112 has not been detecting
```python
async def foo():
        await nursery.start(bar)
```
and only worked properly on `start_soon`.


Zac-HD (approved):
I like this as clean improvement over the status quo; happy to take cleanups in a future PR if that works for you?

Also doesn't looks like this includes a fix for the `await nursery.start(...)` issue you mentioned; if I'm right either fix here or open a tracking issue?

jakkdl:
oops, forgot to push the 112 stuff

PR #242, opened 2024-05-03
ASYNC912: timeout/cancelscope with only conditional checkpoints

jakkdl:
oh, turns out #183 was already planning on using ASYNC912.

It looks like I broke some stuff, but I gotta go for the day so I thought I'd push in case you have time to review `tests/eval_files/async912.py` and have any opinions on the behavior in some of the edge cases in there.


tests/eval_files/async912.py
```diff
@@ -0,0 +1,108 @@
...
+    with trio.move_on_after(0.1):  # error: 4
+        ...
+    with trio.move_on_at(0.1):  # error: 4
+        ...
+    with trio.fail_after(0.1):  # error: 4
+        ...
+    with trio.fail_at(0.1):  # error: 4
+        ...
+    with trio.CancelScope(0.1):  # error: 4
+        ...
```
Zac-HD, line 15: Feels like we should just cover the `CancelScope` case with `ASYNC100` and reserve this optional check for the cases where the await is conditional.

It seems like a semantically distinct problem: "this timeout can't fire" vs "you might be unable to cancel this code like you expected to"... but Trio says that no, these are actually the same thing.  We should be really clear about that in the docs.

tests/eval_files/async912.py
```diff
@@ -0,0 +1,108 @@
...
+
+    with open(""):
+        ...
```
Zac-HD, line 18: add a comment to explain?  ('check that non-cancel-scope context managers are unaffected', or something)

tests/eval_files/async912.py
```diff
@@ -0,0 +1,108 @@
...
+
+    with trio.move_on_after(0.1):
+        await trio.lowlevel.checkpoint()
+
+    with trio.move_on_after(0.1):  # error: 4
+        with trio.move_on_after(0.1):  # error: 8
+            ...
+
+    with trio.move_on_after(0.1):  # TODO: should probably raise an error?
+        with trio.move_on_after(0.1):
```
Zac-HD, line 28: Because the inner `move_on_*` swallows the cancellation?

I think this would be prone to false alarms, because it depends on which one triggers first (if the outer, we're fine, for example).  I'm concerned that trying to be fancy would disallow tricky-but-valid cases, so let's just leave this one alone and accept some missed alarms.

tests/eval_files/async912.py
jakkdl: This turned out to be real tricky to fix properly - but ASYNC100 is now fully handled by Visitor91x so as not to have its logic duplicated. That surfaced a bunch of issues, but those are all handled.

jakkdl:
Got the major shit working, but needs some cleanup, comments, documentation and a self-review pass. So no need to review yet.

tests/eval_files/async912.py
```diff
@@ -0,0 +1,157 @@
...
+    with trio.move_on_after(0.1):
+        await trio.lowlevel.checkpoint()
+
+    with trio.move_on_after(0.1):  # ASYNC912: 9
+        with trio.move_on_after(0.1):  # ASYNC912: 13
+            if bar():
+                await trio.lowlevel.checkpoint()
+
+    # We don't know which cancelscope will trigger first, so to avoid false
+    # positives on tricky-but-valid cases we don't raise any error for the outer one.
```
Zac-HD, line 63: nit, b/c clear in this case: I prefer "false alarm" vs "missed alarm", which is immediately descriptive.

By comparison 'false positive' vs 'false negative' is easy to confuse (is detecting a problem positive?), although still better than the abuse 'Type {1,2,I,II} error' terminology which gives absolutely no hints to the listener.

Zac-HD (approved):
Looks good!  

Happy to merge as-is and follow up on comments later if you want.

tests/eval_files/async912.py
jakkdl: Haha, I've seen you point this out about a dozen times and keep misremembering/forgetting which way to do it :D

Fixing

flake8_async/visitors/helpers.py
```diff
@@ -335,12 +348,30 @@ def with_has_call(
...
+            assert isinstance(res["library"], (cst.Name, cst.Attribute))
             assert isinstance(res["function"], cst.Name)
-            if res["library"].value not in base:
+            library_node = res["library"]
+            for library_str in base:
+                if (
+                    isinstance(library_node, cst.Name)
+                    and library_str == library_node.value
+                ):
+                    break
```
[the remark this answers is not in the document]
jakkdl: yes, though sounds complicated to engineer. I'll give it a try and see if I get anywhere easily, otherwise throw in a `TODO` and maybe deal with it at a later point.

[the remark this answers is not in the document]
jakkdl: Ended up being straightforward to write, but got sidetracked by https://github.com/Instagram/LibCST/issues/1143

jakkdl:
Added a couple corner cases, I don't think wrapping calls is terribly common - but if it is feel free to open an issue and I'll make it search for the matching calls differently.
```python
    # wrapped calls do not raise errors
    with customWrapper(trio.fail_at(10)):
        ...
    with (res := trio.fail_at(10)):
        ...
    # but saving with `as` does
    with trio.fail_at(10) as res:  # ASYNC912: 9
        if bar():
```

PR #244, opened 2024-05-13
add asyncio/anyio taskgroup support to ASYNC101

### LINKED ISSUE #215: [anyio/asyncio].TaskGroup support for 101, 111, 112

jakkdl:
* add asyncio/anyio taskgroup support to ASYNC101. Fixes #215
* Fix `func_has_decorator` to recognize decorators that are calls but not attributes.
* Clarify difference between ASYNC101 and ASYNC119 in docs. (see https://github.com/python-trio/flake8-async/issues/211#issuecomment-2088481877)


Zac-HD (approved):
10/10, no notes.

(but literally, I just read straight through and didn't have any comments.  Thanks again for your ongoing great work on this!)

jakkdl:
> 10/10, no notes.
> 
> (but literally, I just read straight through and didn't have any comments. Thanks again for your ongoing great work on this!)

haha, new achievement unlocked! Thank you ^^

Issue #227, opened 2024-03-29
Add new `ASYNC912` rule to warn on missing checkpoint-or-exceptions


Zac-HD:
```python
import contextlib, trio

@trio.run
async def livelocks():
    with trio.move_on_after(0.1):  # can't cancel, because sleep() raises before checkpointing
        while True:
            try:
                await trio.sleep("1")
            except TypeError:
                pass
            # or
            with contextlib.suppress(TypeError):
                await trio.sleep("1")
```

Per https://trio.readthedocs.io/en/stable/reference-core.html#checkpoints, Trio's own functions unconditionally checkpoint _unless_ an exception is raised, in which case they may or may not checkpoint.  Glossing the design goals of `flake8-async` as (1) enforce these conditions in your own code and (2) miscellaneous, to date we've just ignored the possible lack of checkpoints when an exception is raised.  In fairness this is a pretty rare pattern; I think it's only bitten me twice at work - though once very badly.

I think the ideal situation here is to add a new `ASYNC912` rule, with a corresponding `async912-context-managers=` config option to expand the denylist of context managers presumed to catch exceptions without re-raising them.  I think assuming that _any_ context manager might discard an exception is just too noisy for realistic use, although you could set `=*` to emulate that if you really wanted to.

How to solve this: add `await trio.lowlevel.checkpoint()` to the loop body, _outside_ the try/except or the with block.  An autofix would be nice eventually, but this is rare enough that I wouldn't bother unless it's really easy.

jakkdl:
I think the best way of implementing this is to further extend `Visitor91X` to handle checkpoints in contexts-with-suppressed-exceptions in some special way. Easiest would be to fully ignore them (if some option is set) and then raise 910/911 as normal, but it should be doable to set some special marker such that 912 is raised in those cases instead.

Given that we're piggybacking on 910/911, it should be easy to piggyback on their autofixing as well.

jakkdl:
Oh wait no, the try/except version is already supported by 910/911:
```python
async def no_checkpoint():
    try:
        await trio.sleep("1")
    except ValueError:
        ...
```
gives an ASYNC910 warning. So there's two separate feature requests:
1. Add an option for async910/911 for exception-eating context managers. Possibly raising a different code when the only checkpoint is one whose checkpoint is not guaranteed due to user-supplied context manager.
2. Integrate ASYNC100 into Visitor91x, to also make use of e.g.
```python
async def does_not_currently_raise_async100():
[...]

PR #255, opened 2024-05-24
Add ASYNC913, indefinite-loop-no-guaranteed-checkpoint. Fix async91x bugs

### LINKED ISSUE #240: New rule: no guaranteed checkpoint in infinite loop
```python
# this raises 910
async def no_checkpoint():
    if ...:
        await trio.sleep("1")
# and this
async def no_checkpoint():
    while bar():
        if ...:
            await trio.sleep("1")
# but not this
async def no_checkpoint():
    while True:
        if ...:
            await trio.sleep("1")
```
same for try/except.

The problem likely stems from the fact that the function never returns, and as such it never returns *without checkpointing* - which is where 910 would trigger. Found it while exploring #227, and would be particularly problematic if/when expanding ASYNC100 to use Visitor91x.

jakkdl:
Adds a new check ASYNC913, indefinite-loop-no-guaranteed-checkpoint.

Will likely have some minor merge conflicts with #254 
Fixes #240.
Together with #254 it fixes the example given in #227. The error will be for the loop though, and not the context manager.

I also noticed a few bugs when implementing it, that also are fixed
* ASYNC911 crashed when trying to autofix
```python
while True: yield
```
because the autofix assumed the loop body was an `IndentedBlock`.
* ASYNC911 didn't add a library import when using InsertCheckPointsInLoopBody, i.e. when a checkpoint was needed due to the artificial statement injected at start of loop body. See eval_files/async911_insert_library.py
* ASYNC910 didn't add a library import when adding a checkpoint at the end of a function body. (as opposed to before a `return`). See eval_files/async910_insert_library.py

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,7 +4,7 @@ Changelog
 
 *[CalVer, YY.month.patch](https://calver.org/)*
 
-24.5.3
+24.5.4
 ======
 - Add ASYNC913: Indefinite loop with no guaranteed checkpoint.
 - Fix bugs in ASYNC910 and ASYNC911 autofixing where they sometimes didn't add a library import.
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -6,6 +6,12 @@ Changelog
 
 24.5.3
 ======
+- Add ASYNC913: Indefinite loop with no guaranteed checkpoint.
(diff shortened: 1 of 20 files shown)


Zac-HD (approved):
All looks good to me!

I'll leave this for you to merge after #254 (so we release with both); and make sure to close #227 which isn't yet autolinked.

PR #256, opened 2024-05-27
add ASYNC300 create-task-no-reference

### LINKED ISSUE #207: Checks for `asyncio.create_task()`
via https://github.com/Zac-HD/flake8-async-archive/issues/5: the docs for [asyncio.create_task](https://docs.python.org/3.11/library/asyncio-task.html#asyncio.create_task) say:

> Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks. A task that isn’t referenced elsewhere may get garbage collected at any time, even before it’s done.

It's really easy to miss this and try to fire-and-forget. It would be great if flake8-async detected this.

-------------

I also wonder whether we should have an optional check which recommends using `TaskGroup` instead of `asyncio.create_task`, `loop.run_until_complete`, etc, but on balance think it's better to have documentation explaining why you should use `anyio.TaskGroup` instead (e.g. better cancellation semantics as part of structured concurrency).

jakkdl:
Fixes #207, originally requested by @alicederyn in https://github.com/Zac-HD/flake8-async-archive/issues/5

The doc and error message can probably be improved.

I went for quite comprehensive checks for possible ways of not using the value. I don't think ~anybody that installs this plugin would hit any of them, but if/when this gets packaged into ruff I could imagine one of the weirder corner cases getting hit.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -6,7 +6,7 @@ Changelog
 
 24.5.5
 ======
-- Add :ref:`ASYNC120 <async120>` create-task-no-reference
+- Add :ref:`ASYNC300 <async300>` create-task-no-reference
 
 24.5.4
 ======
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,13 +4,17 @@ Changelog
 
 *[CalVer, YY.month.patch](https://calver.org/)*
 
+24.5.5
+======
+- Add :ref:`ASYNC120 <async120>` create-task-no-reference
+
 24.5.4
 ======
-- Add ASYNC913: Indefinite loop with no guaranteed checkpoint.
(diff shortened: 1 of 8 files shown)


jakkdl:
whoa, slow tests found a crash!
Somehow the `value` attribute of a `cst.Attribute` is neither `cst.Attribute` nor `cst.Name`. [The docs](https://libcst.readthedocs.io/en/latest/nodes.html#libcst.Attribute) *do* say that it can be any `BaseExpression`, but I'm failing to make sense of that. Time to dig into the mccabe==0.7.0 source code

Zac-HD (approved):
Renumbered to `ASYNC300` following the scheme I just proposed in #257, looks great!

jakkdl:
and your changes look good too, thanks!

PR #259, opened 2024-05-28
make async913 disabled by default

jakkdl:
oopsies

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,10 @@ Changelog
 
 *[CalVer, YY.month.patch](https://calver.org/)*
 
+24.5.6
+======
+- Make :ref:`ASYNC913 <async913>` disabled by default, as originally intended.
+
 24.5.5
 ======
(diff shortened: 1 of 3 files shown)


Zac-HD (approved):
Oops, should have noticed in review 😅

jakkdl:
...although looking at it more closely, we already default `--disable` to `ASYNC9`, and what `disable_codes_by_default` achieves is adding the codes to `--ignore`. So in effect we're doubly disabling the codes and you have to both `--select` and `--enable` them to get run? Seems bad, we probably shouldn't bother with calling 
...nvm nope! Turns out we change the default to `--disable` depending on if we're standalone or not.

This is a jolly mess, which maaybe made sense before flake8 started enforcing three-letter codes in `--ignore` (and presumably `--select`). We probably should stop calling `option_manager.extend_default_ignore(default_disabled_error_codes)` and make `--disable` always have the default of `ASYNC9`, and lean into that end-users should only use enable/disable, and never try to use select/ignore.

jakkdl:
should've marked it draft while writing up that comment x)

PR #264, opened 2024-05-29
various small fixes

jakkdl:
Various small fixes I noticed while going through code/doc/issues.

* doc updates to 113 and `startable-in-context-manager`
* ASYNC102 await-in-finally-or-cancelled no longer treats `open_nursery` as a cancel scope.
  * While you can make the cancelscope inside the nursery shielded with a timeout, that was never actually supported (there was no logic for seeing `nursery.cancelscope.shield = True`, only for `nursery.shield = True`, which is not possible)
  * `open_nursery` is itself an async function so you can't safely open a nursery anyway.
*  async112 error message now specifies if it's a nursery or taskgroup.
   * for dumb reasons I confused myself into thinking async111 & async112 didn't have asyncio support, and fully rewrote that from scratch :upside_down_face: But some of the stuff was better so I kept it after noticing my mistake and rebasing onto an updated main.

### DIFF
--- a/docs/rules.rst
+++ b/docs/rules.rst
@@ -55,9 +55,9 @@ ASYNC112 : useless-nursery
 
 _`ASYNC113` : start-soon-in-aenter
     Using :meth:`~trio.Nursery.start_soon`/:meth:`~anyio.abc.TaskGroup.start_soon` in ``__aenter__`` doesn't wait for the task to begin.
-    This will only warn about built-in functions and those listed in _`ASYNC114`.
-    If you're starting a function that does not define `task_status`, then neither will trigger.
     Consider replacing with :meth:`~trio.Nursery.start`/:meth:`~anyio.abc.TaskGroup.start`.
+    This will only warn about functions listed in :ref:`ASYNC114 <async114>` or known from Trio.
+    If you're starting a function that does not define `task_status`, then neither will trigger.
 
 _`ASYNC114` : startable-not-in-config
     Startable function (i.e. has a ``task_status`` keyword parameter) not in :ref:`--startable-in-context-manager <--startable-in-context-manager>` parameter list, please add it so ASYNC113 can catch errors when using it.
--- a/docs/rules.rst
+++ b/docs/rules.rst
@@ -55,6 +55,8 @@ ASYNC112 : useless-nursery
 
 _`ASYNC113` : start-soon-in-aenter
     Using :meth:`~trio.Nursery.start_soon`/:meth:`~anyio.abc.TaskGroup.start_soon` in ``__aenter__`` doesn't wait for the task to begin.
+    This will only warn about built-in functions and those listed in _`ASYNC114`.
+    If you're starting a function that does not define `task_status`, then neither will trigger.
     Consider replacing with :meth:`~trio.Nursery.start`/:meth:`~anyio.abc.TaskGroup.start`.
 
 _`ASYNC114` : startable-not-in-config
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -265,8 +265,9 @@ Example
 ``startable-in-context-manager``
 --------------------------------
 
-Comma-separated list of methods which should be used with :meth:`trio.Nursery.start`/:meth:`anyio.abc.TaskGroup.start` when opening a context manager,
-in addition to the default :func:`trio.run_process`, :func:`trio.serve_tcp`, :func:`trio.serve_ssl_over_tcp`, and :func:`trio.serve_listeners`.
+Comma-separated list of methods which should be used with :meth:`trio.Nursery.start`/:meth:`anyio.abc.TaskGroup.start` when opening a context manager.
+This includes startable functions in the trio and anyio standard library by default, namely :func:`trio.run_process`, :func:`trio.serve_tcp`, :func:`trio.serve_ssl_over_tcp`, :func:`trio.serve_listeners`, :func:`trio.serve`, :func:`anyio.run_process`, :func:`anyio.serve_tcp`, :func:`anyio.serve_ssl_over_tcp`, :func:`anyio.serve_listeners`, and :func:`anyio.serve`.
+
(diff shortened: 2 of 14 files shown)


jakkdl:
Don't think this needs a version/changelog bump, but it would probably be good to have some check for if the changelog is updated in a pull request as that's been missed in the past.

docs/usage.rst
```diff
@@ -265,8 +265,9 @@ Example
 ``startable-in-context-manager``
 --------------------------------
 
-Comma-separated list of methods which should be used with :meth:`trio.Nursery.start`/:meth:`anyio.abc.TaskGroup.start` when opening a context manager,
-in addition to the default :func:`trio.run_process`, :func:`trio.serve_tcp`, :func:`trio.serve_ssl_over_tcp`, and :func:`trio.serve_listeners`.
+Comma-separated list of methods which should be used with :meth:`trio.Nursery.start`/:meth:`anyio.abc.TaskGroup.start` when opening a context manager.
+This includes startable functions in the trio and anyio standard library by default, namely :func:`trio.run_process`, :func:`trio.serve_tcp`, :func:`trio.serve_ssl_over_tcp`, :func:`trio.serve_listeners`, :func:`trio.serve`, :func:`anyio.run_process`, :func:`anyio.serve_tcp`, :func:`anyio.serve_ssl_over_tcp`, :func:`anyio.serve_listeners`, and :func:`anyio.serve`.
```
Zac-HD, line 269: I think that the terms 'standard library' and 'by default' can be confusing here; I'd say something like "We then add known functions from AnyIO and Trio to this list, specifically [...]."

```
docs/usage.rst:268: WARNING: py:func reference target not found: trio.serve
docs/usage.rst:268: WARNING: py:func reference target not found: anyio.serve_tcp
docs/usage.rst:268: WARNING: py:func reference target not found: anyio.serve_ssl_over_tcp
docs/usage.rst:268: WARNING: py:func reference target not found: anyio.serve_listeners
docs/usage.rst:268: WARNING: py:func reference target not found: anyio.serve
docs/usage.rst:271: WARNING: undefined label: 'async114'
```

flake8_async/visitors/visitors.py
```diff
@@ -113,19 +113,22 @@ def visit_With(self, node: ast.With | ast.AsyncWith):
...
+                item.context_expr, "create_task_group", base="anyio"
+            ):
+                nursery_type = "taskgroup"
+            # check for asyncio.TaskGroup
+            elif get_matching_call(item.context_expr, "TaskGroup", base="asyncio"):
+                nursery_type = "taskgroup"
                 start_methods = ("create_task",)
+            else:
+                assert True
+                continue
```
Zac-HD, line 131: I'd just put a `# pragma: no cover  # see coverage issue #nnnn` here

tests/eval_files/async102.py
```diff
@@ -111,7 +111,9 @@ async def foo():
         myvar = True
         with trio.open_nursery(10) as s:
             s.shield = myvar
-            await foo()  # safe in theory, error: 12, Statement("try/finally", lineno-6)
+            # this is not safe in theory - because `trio.open_nursery` is an async cm,
+            # so it's not possible to open a nursery at all.
+            await foo()  # error: 12, Statement("try/finally", lineno-8)
```
Zac-HD, line 116: So this should be `async with`, yeah?  Also nurseries only checkpoint on exit, so I think it _is_ safe?

docs/rules.rst
```diff
@@ -55,6 +55,8 @@ ASYNC112 : useless-nursery
 
 _`ASYNC113` : start-soon-in-aenter
     Using :meth:`~trio.Nursery.start_soon`/:meth:`~anyio.abc.TaskGroup.start_soon` in ``__aenter__`` doesn't wait for the task to begin.
+    This will only warn about built-in functions and those listed in _`ASYNC114`.
```
Zac-HD, line 58:
```suggestion
    This will only warn about functions listed in _`ASYNC114` or known from Trio or AnyIO.
```

tox.ini
```diff
@@ -18,7 +18,9 @@ deps =
     hypothesmith >= 0.3.3
     trio
 commands =
-    pytest {posargs:-n auto}
+    coverage erase
+    coverage run -m pytest {posargs:-n auto}
+    coverage html
```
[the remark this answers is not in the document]
jakkdl: sorry, that commit was merely to make it easier to reproduce and rule out pytest-cov. I did not mean for the changes to actually get merged

docs/usage.rst
jakkdl: Oh, turns out there is not a single function in anyio that takes a `task_status` parameter. The `serve_xxx` functions don't exist, and neither [`anyio.run_process`](https://anyio.readthedocs.io/en/stable/api.html#anyio.run_process) has a `task_status` parameter nor any of the `serve` methods do. A search through the code and docs also reveal nothing. So ASYNC113 in fact does have several false alarms currently. Should change the code to also take into account the base of the attribute and not match on any `xxx.run_process`.

Zac-HD (approved):
LGTM overall; since this won't trigger a release let's deal with the false alarms in a follow-up 🙂

tests/eval_files/async102.py
jakkdl: wait, it doesn't block/checkpoint on exit either??

```python
import trio

async def foo():
    with trio.CancelScope() as cs:
        cs.cancel()
            print("in nursery")
        print("after nursery")

trio.run(foo)
```

```sh
python foo.py
in nursery
after nursery
```

jakkdl: > It does not block on entry; on exit it blocks until all child tasks have exited.

oh, hm. I guess that is meant to be read "if there are child tasks still alive, block until they have exited. If there are no child tasks, or they have all exited, it doesn't block."
Which in the end means that `open_nursery` never checkpoints on entry, and *sometimes* checkpoints on exit.

Issue #262, opened 2024-05-28
Add lint error for _any_ `except` block that might raise


Zac-HD:
c.f. https://github.com/python-trio/trio/issues/455#issuecomment-2135987009 - this would be similar to `ASYNC102` but I think ideally a separate rule so that people who wish to disable this (if lost exceptions aren't a big deal for them) don't reintroduce lost _cancellations_.

We should probably also do something similar for `__aexit__` methods, although this gets increasingly awkward for `@asynccontextmanager` functions and so that should probably be a further separate rule.

jakkdl:
`__aexit__` seems straightforward, we just treat the content of the whole method as if it's a `finally:`, and can probably use `ASYNC102` for it.
For `@asynccontextmanager`, 102/120 should already cover if the yield is inside a `try:` - so I think that should maybe have an entirely separate rule for `yield-in-async-cm-not-in-try`?

jakkdl:
with #307 closed this is now fully addressed once #306 is merged

PR #265, opened 2024-05-29
add async120, await-in-except

jakkdl:
First part of #262 

It's maybe not great to have the same error messages for both 102 and 120, and docs could maybe be improved as well.

### DIFF
--- a/docs/rules.rst
+++ b/docs/rules.rst
@@ -77,7 +77,7 @@ _`ASYNC119` : yield-in-cm-in-async-gen
 
 _`ASYNC120` : await-in-except
     Dangerous :ref:`checkpoint` inside an ``except`` block.
-    If this checkpoint is cancelled, the current active exception will be replaced by the ``Cancelled`` exception, and cannot be reraised later. 
+    If this checkpoint is cancelled, the current active exception will be replaced by the ``Cancelled`` exception, and cannot be reraised later.
     This will not trigger when :ref:`ASYNC102 <ASYNC102>` does, and if you don't care about losing non-cancelled exceptions you could disable this rule.
     This is currently not able to detect asyncio shields.
 
--- a/docs/rules.rst
+++ b/docs/rules.rst
@@ -75,7 +75,7 @@ _`ASYNC119` : yield-in-cm-in-async-gen
    ``yield`` in context manager in async generator is unsafe, the cleanup may be delayed until ``await`` is no longer allowed.
    We strongly encourage you to read `PEP 533 <https://peps.python.org/pep-0533/>`_ and use `async with aclosing(...) <https://docs.python.org/3/library/contextlib.html#contextlib.aclosing>`_, or better yet avoid async generators entirely (see `ASYNC900`_ ) in favor of context managers which return an iterable :ref:`channel/stream/queue <channel_stream_queue>`.
 
-ASYNC120 : await-in-except
+_`ASYNC120` : await-in-except
     ``await`` inside ``except`` must have shielded :ref:`cancel scope <cancel_scope>` with timeout.
     If not, the async call might get cancelled, suppressing the exception that was caught.
     This will not trigger when :ref:`ASYNC102 <ASYNC102>` does, and if you don't care about losing non-cancelled exceptions you could disable this rule.
--- a/docs/rules.rst
+++ b/docs/rules.rst
@@ -76,8 +76,8 @@ _`ASYNC119` : yield-in-cm-in-async-gen
    We strongly encourage you to read `PEP 533 <https://peps.python.org/pep-0533/>`_ and use `async with aclosing(...) <https://docs.python.org/3/library/contextlib.html#contextlib.aclosing>`_, or better yet avoid async generators entirely (see `ASYNC900`_ ) in favor of context managers which return an iterable :ref:`channel/stream/queue <channel_stream_queue>`.
 
 _`ASYNC120` : await-in-except
-    ``await`` inside ``except`` must have shielded :ref:`cancel scope <cancel_scope>` with timeout.
-    If not, the async call might get cancelled, suppressing the exception that was caught.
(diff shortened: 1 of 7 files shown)


docs/rules.rst
```diff
@@ -73,6 +75,12 @@ _`ASYNC119` : yield-in-cm-in-async-gen
    ``yield`` in context manager in async generator is unsafe, the cleanup may be delayed until ``await`` is no longer allowed.
    We strongly encourage you to read `PEP 533 <https://peps.python.org/pep-0533/>`_ and use `async with aclosing(...) <https://docs.python.org/3/library/contextlib.html#contextlib.aclosing>`_, or better yet avoid async generators entirely (see `ASYNC900`_ ) in favor of context managers which return an iterable :ref:`channel/stream/queue <channel_stream_queue>`.
 
+_`ASYNC120` : await-in-except
+    ``await`` inside ``except`` must have shielded :ref:`cancel scope <cancel_scope>` with timeout.
+    If not, the async call might get cancelled, suppressing the exception that was caught.
+    This will not trigger when :ref:`ASYNC102 <ASYNC102>` does, and if you don't care about losing non-cancelled exceptions you could disable this rule.
+    This is currently not able to detect asyncio shields.
```
Zac-HD, line 82: I don't think I want to suggest shielding for this - we don't want to render it impossible to cancel while in these clauses, just warn that the current exception might be discarded.

(Of course, if it _is_ shielded then that should suppress the error message.  Maybe this is a more general pattern, where checkpoints inside a shielded cancel scope "shouldn't count"?  But that's a future issue)

```suggestion
    Dangerous :ref:`checkpoint` inside an ``except`` block.
    If this checkpoint is cancelled, the current active exception will be replaced by the ``Cancelled`` exception, and cannot be reraised later. 
    This will not trigger when :ref:`ASYNC102 <ASYNC102>` does, and if you don't care about losing non-cancelled exceptions you could disable this rule.
    This is currently not able to detect asyncio shields.
```

flake8_async/visitors/visitor102.py
```diff
@@ -23,6 +23,10 @@ class Visitor102(Flake8AsyncVisitor):
             "await inside {0.name} on line {0.lineno} must have shielded cancel "
             "scope with a timeout."
         ),
+        "ASYNC120": (
+            "await inside {0.name} on line {0.lineno} must have shielded cancel "
+            "scope with a timeout."
```
Zac-HD, line 28:
```suggestion
            "checkpoint inside {0.name} on line {0.lineno} will discard the active "
            "exception if cancelled."
```

flake8_async/visitors/visitor102.py
```diff
@@ -60,7 +64,11 @@ def async_call_checker(
         if self._critical_scope is not None and not any(
             cm.has_timeout and cm.shielded for cm in self._trio_context_managers
         ):
-            self.error(node, self._critical_scope)
+            if self._critical_scope.name == "except":
+                error_code = "ASYNC120"
+            else:
+                error_code = "ASYNC102"
+            self.error(node, self._critical_scope, error_code=error_code)
```
Zac-HD, line 71: I think the critical scopes are only those which might have a `CancelledError` active, and we want to apply this to _all_ (unshielded) `except` blocks.

jakkdl: This is just a matter of whether you want both of 102 and 120 to be raised for checkpoints in critical scopes.
You might be getting confused by "except" being a ~magical constant to indicate an except is non-critical? I can add comments to further clarify that, and/or restructure a bunch of things to use a different magical word. The reason for it is it being reused for formatting the error message.

jakkdl:
I noticed the "exclude x.aclose()" just now, I don't remember the reasoning behind that, so idk if 120 should respect it as well.

jakkdl:
I just reread https://github.com/python-trio/trio/issues/455#issuecomment-2136041817 and noticed
>  I guess we should add any other except-that-raises here.

and realized that this check does not filter on whether-the-except-would-otherwise-raise, so this will currently raise:
```python
try:
  ...
except ValueError:
  # we don't give a shit about ValueError
  await doSomeOtherThing()
```

can fix it by saving the potential async120 errors in a container and only report them upon reaching a `raise`

tests/eval_files/async102.py
```diff
@@ -265,3 +266,20 @@ async def foo_nested_async_for():
...
+
+        async def foobar():
+            await foo()
+
+
+async def foo_nested_funcdef():
+    try:
+        ...
+    except:
+        x = lambda: await foo()
```
Zac-HD, line 285: `SyntaxError: 'await' outside async function` - there's no such thing as an async lambda.

We might want to check that our test files are syntactically valid to avoid any subtle issues in future?

flake8_async/__init__.py
```diff
@@ -38,7 +38,7 @@
 
 
 # CalVer: YY.month.patch, e.g. first release of July 2022 == "22.7.1"
-__version__ = "24.5.6"
+__version__ = "24.5.7"
```
Zac-HD, line 41:
```suggestion
__version__ = "24.6.1"
```

since I've taken so long 😅

Zac-HD (approved):
Looks good, thanks for your patience!  Merge at your discretion.

tests/eval_files/async102.py
jakkdl: After testing, it appears that [`ast.parse`](https://docs.python.org/3/library/ast.html#ast.parse) does not pick up on it - and neither mypy nor ruff has any checks for it. Calling [`compile`](https://docs.python.org/3/library/functions.html#compile) on it does surface the error, as it does additional scoping checks that `ast.parse` does not do.

jakkdl: (more obvious errors have been caught in the past by mypy/ruff). Opened #268

PR #279, opened 2024-08-07
Add config option for async900 decorators

### LINKED ISSUE #277: Allow user-configured decorator names to disable `ASYNC900`
Currently, we recognize that decorators named `@asynccontextmanager` and `@fixture` (including attributes and calls, so e.g. `@pytest.fixture(...)` also works) mean the function isn't _really_ an async generator:

https://github.com/python-trio/flake8-async/blob/bae7ab7e3e5163b7a5d0d7966669e58ff533336e/flake8_async/visitors/visitors.py#L415-L418

I think we should make this list extensible, similar to our existing [`no-checkpoint-warning-decorators` list](https://flake8-async.readthedocs.io/en/latest/usage.html#no-checkpoint-warning-decorators) config.  Could be called something like `not-a-generator-decorators`.

The motivation is that you might well want to use the nice syntax of `yield`, which is safe if wrapped in something like https://github.com/groove-x/trio-util/pull/9 (see https://github.com/python-trio/trio/issues/638 for design discussion).  I'm inclined to see if we can upstream something similar into Trio and AnyIO, but even then there will be people on older versions of the libraries (or asyncio!) for years to come and so even if we add to the baseline list later having a config option seems useful.

Zac-HD:
Fixes #277.

### DIFF
--- a/tests/eval_files/async900.py
+++ b/tests/eval_files/async900.py
@@ -3,7 +3,7 @@
 from contextlib import asynccontextmanager
 
 
-async def foo1():  # ASYNC900: 0
+async def foo1():  # ASYNC900: 0, 'asynccontextmanager, fixture, this_is_like_a_context_manager'
     yield
     yield
 
@@ -15,7 +15,7 @@ async def foo2():
 
 @asynccontextmanager
 async def foo3():
-    async def bar():  # ASYNC900: 4
+    async def bar():  # ASYNC900: 4, 'asynccontextmanager, fixture, this_is_like_a_context_manager'
         yield
 
     yield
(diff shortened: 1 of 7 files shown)


tests/eval_files/async900.py
```diff
@@ -54,3 +54,12 @@ async def cm():
 async def another_non_generator():
     def foo():
         yield
+
+
+# ARG --transform-async-generator-decorators=this_is_like_a_context_manager
```
jakkdl, line 59: Minor: I'd probably have this at the top of the file for readability (and in all other files I've had them at the top), esp as it pops up in the other error messages.

jakkdl (commented):
Looks good! :)

PR #282, opened 2024-08-27
add async121 control-flow-in-taskgroup

jakkdl:
closes #261 

Not sure whether to name the rule `control-flow-in-taskgroup`, `control-flow-in-nursery`, or `control-flow-in-taskgroup-or-nursery`. All previous nursery/taskgroup-related rules have in fact been about the cancelscope within them, such that none of them had to decide.
This made me want to double-check that this rule *is* exclusive to nursery/taskgroups, and that the rule shouldn't be expanded to also trigger on control flows inside cancelscopes, but I cannot personally come up with anything at least.

The test isn't *fully* exhaustive with nested loops and stuff, but the visitor is simple enough I think it's unlikely something would sneak through.

PR #285, opened 2024-09-06
Rst linting

jakkdl:
Fixes https://github.com/python-trio/flake8-async/pull/282#discussion_r1745922270
Adds https://pypi.org/project/sphinx-lint/ and fixes all errors pointed out by it.
Also fixing a broken bullet list I manually noticed

I did not add any of the following linters, but that could be revisited

* https://github.com/rstcheck/rstcheck which pointed out the markdown url in `changelog.rst`, but it was otherwise very noisy and didn't point out any actual errors.

* https://github.com/PyCQA/doc8 complains about a ton of lines being too long, so if we want to stick with a max char line length it's the tool to go for. I personally feel it's got much less of an upside compared to code, and can be a major hassle to reflow lines after editing them instead of just letting autowrap handle it.

* https://github.com/twolfson/restructuredtext-lint appears to have very few and very specific rules (that I'm guessing are mostly covered by one or more of the above linters)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -93,3 +93,8 @@ repos:
       hooks:
           - id: pretty-format-toml
             args: [--autofix]
+
+    - repo: https://github.com/sphinx-contrib/sphinx-lint
+      rev: v0.9.1
+      hooks:
+          - id: sphinx-lint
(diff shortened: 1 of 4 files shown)


Zac-HD (approved):
Nice!

PR #287, opened 2024-09-09
ASYNC113 & ASYNC121 false alarm with nested sync functions

### LINKED ISSUE #286: `ASYNC121` false alarm in nested async function definition
```python
import trio

async def fn():

        def foo():
            return  # safe, no lint warning

        async def bar():
            return  # false alarm ASYNC121
```

follows #261, #282; should be a pretty quick fix

jakkdl:
Fixes #286 

I also did a search for `visit_FunctionDef` and `visit_AsyncFunctionDef` in the codebase and caught the same error for async113 (though it should very rarely occur for 113). At first glance a couple other visitors had the same problem, until I noticed those were cst visitors (which instead has an `asynchronous` attribute).

For 113 we could consider having the logic for `visit_FunctionDef` being the same as `visit_AsyncFunctionDef` instead of always setting `self.aenter = False` in case users are transforming sync methods to async methods, but given such exotic stuff we're maybe better off staying away and avoiding super obscure false positives.

We could consider adding a test that checks if any AST visitors only has one of the visitors defined, and could do the same for `With`/`AsyncWith` and `For`/`AsyncFor` (the latter which was in fact only caught in code review for #282). It will have a decent false-positive ratio though, but not very troublesome to have an ignore list or declare dummy methods.

Another approach is preferring cst over AST, and I've been inconsistent when creating new visitors despite https://github.com/python-trio/flake8-async/issues/124. I should try and compile pros/cons and write a comment over there.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,11 @@ Changelog
 
 `CalVer, YY.month.patch <https://calver.org/>`_
 
+24.9.2
+======
+- Fix false alarm in :ref:`ASYNC113 <async113>` and :ref:`ASYNC121 <async121>` with sync functions nested inside an async function.
+
+
 24.9.1
 ======
 - Add :ref:`ASYNC121 <async121>` control-flow-in-taskgroup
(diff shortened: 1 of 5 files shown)


Zac-HD (approved):
Thanks @jakkdl!

Issue #283, opened 2024-08-31
ASYNC102 await-in-finally-or-cancelled false alarms on nurseries


jakkdl:
https://github.com/python-trio/flake8-async/blob/225f15a98db4bb8ce4600127fd0fe90bc5c9c04b/tests/eval_files/async102.py#L117-L125

TODO:
* [ ] Special-case `async with open_nursery(...)`, and `async with anyio.create_task_group(...)` as not being checkpoints. (also affects other rules)
* [ ] handle `[nursery_name].cancel_scope.[deadline/shield]`

jakkdl:
The other rules that could be relevant are:
* ASYNC100 cancel-scope-no-checkpoint
  * A cancel scope with only a nursery can never be cancelled, so this should ignore `open_nursery`/`create_task_group`
* ASYNC120 await-in-except
  * Same as 102, and also already fixed by #289 
* ASYNC91x
  * nurseries can checkpoint, so these should not ignore nurseries, and do not need to be changed.

```python
async def foo(): # should not give ASYNC910
  async with trio.open_nursery() as nursery():
    nursery.start_soon(bar)
```

tl;dr, ASYNC100 also needs changing

jakkdl:
Hrm, neither of these statements are cancellation sources
```python
my_nursery = await trio.open_nursery()
async with my_nursery:
    ...
```
So it's not just `async with trio.open_nursery()` that needs to be checked against.
Both of those cases are definitely less common, but far from unheard of. The former is an easy fix, the latter is a bit more tricky but we're already type-tracking nurseries for other reasons so shouldn't be too bad.

PR #289, opened 2024-09-10
fix various issues with ASYNC102

jakkdl:
Fixes most of #283 and some other things I noticed, only remaining thing being checking other rules for if they should also have special handling of `open_nursery`/`create_task_group`.

Note that `trio.move_on_after(shield=True)` has not been released yet


Zac-HD (approved):
Looks good, thanks @jakkdl!

jakkdl:
given the quick review I'll leave changes to ASYNC100 to another PR :)

PR #292, opened 2024-09-18
add ASYNC122 delayed-entry-of-relative-cancelscope

### LINKED ISSUE #290: move_on_after / fail_after not immediately entered
See https://github.com/python-trio/trio/issues/2512 and https://github.com/python-trio/trio/pull/3010

The PR discussion suggests gating it on currently running Trio version, I personally think that's too prone to missed errors and should instead always warn - but prominently mention that you should disable the check if you only support new Trio versions.

jakkdl:
fixes #290 

It doesn't handle the complex case of
```python
foo = trio.fail_after
bar = foo(5)
trio.sleep(3)
with bar:
  ...
```
and while it would be possible to error anytime somebody refers to `trio.fail_after` not directly inside a `with` call it'd probably mostly just lead to unnecessary false positives.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,10 @@ Changelog
 
 `CalVer, YY.month.patch <https://calver.org/>`_
 
+24.9.4
+======
+- Add :ref:`ASYNC122 <async122>` delayed-entry-of-relative-cancelscope.
+
 24.9.3
 ======
 - :ref:`ASYNC102 <async102>` and :ref:`ASYNC120 <async120>`:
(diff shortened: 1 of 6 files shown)


jakkdl:
The error message could perhaps mention anyio versions as well, in case @agronholm plans on incorporating https://github.com/python-trio/trio/pull/3010 soonish. But regardless it'll give errors for `anyio.move_on_after` and `anyio.fail_after`.

Zac-HD (approved):
Looks good, thanks @jakkdl!

PR #307, opened 2024-10-25
add async124 yield-in-asynccm-not-in-try

jakkdl:
last part of #262, though I'm not 100% on whether this should be a rule in flake8-async.

have not bumped version or added to changelog, and will extend the tests if we decide to go ahead.
(and if not make sure the fix to Visitor102.Visit_Try gets merged elsewhere)


Zac-HD:
I'd lean towards "don't bother doing this", since we're already covering the try/finally case I think there aren't any `await`s that wouldn't already be warned-about.

jakkdl:
> I'd lean towards "don't bother doing this", since we're already covering the try/finally case I think there aren't any `await`s that wouldn't already be warned-about.

right, I guess if the `yield` raises then any `await`s can't cause any problems so this is about a completely different problem that's fairly well known and usually quite noticeable when it's not followed.

PR #309, opened 2024-10-29
add async124 async-function-could-be-sync

### LINKED ISSUE #253: New rule: async function with no `await` could be sync
The ASYNC100-version of ASYNC91x, i.e. the trivial case where there is no `await`/`async [for/with]` in a method at all. As opposed to ASYNC91x this could be default-enabled, will be relevant for asyncio users, and should be much less noisy.
We could silence 910 and 911 for the trivial case to mirror async100/async912, or expect users to disable async-func-no-await when enabling 910 & 911.

Examples:
```python
# currently raises 910, would raise async-func-no-await
async def foo():
    print("bar")
# would only raise 910
async def bar():
    if ...:
        await foo()
# async-func-no-await does not care about func vs generator
# raises 911 and async-func-no-await
async def foo_gen():
    yield "hello"
# does not give any warnings
async def foo_always():
    if ...:
        await foo()
    else:
        await bar()
```

Shoutout to @oremanj for giving inspiration for it in gitter.

> I think it is still a good idea for linters to insist that every async function await something, because async functions are more expensive to call than synchronous functions, and if it doesn't await anything then it should just be synchronous. In order to get a checkpoint guarantee you need to await something on every path; this is definitely helpful if you use Trio (or if you're writing an anyio library that people might use on Trio), but it's hard to argue that it provides any meaningful benefit on asyncio when the built-in functions don't follow the rule.

jakkdl:
fixes #253

given that converting an async test to a sync test will give weird errors https://github.com/agronholm/anyio/issues/803 https://github.com/pytest-dev/pytest/issues/10839 we should probably err on the side of not erroring, at least until pytest/anyio is fixed, ~~which would extend to 910 and 911.~~


Zac-HD (commented):
Implementation looks good, but I'm pretty concerned about the false-alarm rate.

- Let's also exclude methods on a class; in those cases it's reasonably likely that it 'has to be async' in order to match the interface of some other type
- If there's a convenient way to try it out we could get some empirical feedback?  I can't think of one though.

jakkdl:
> Implementation looks good, but I'm pretty concerned about the false-alarm rate.
> 
> * Let's also exclude methods on a class; in those cases it's reasonably likely that it 'has to be async' in order to match the interface of some other type

Hmhm, I also feel like this would lose a bunch of the upside of it. Maybe a separate error code?

> * If there's a convenient way to try it out we could get some empirical feedback?  I can't think of one though.

push a release and wait for complaints? :laughing:
We could push it as default-disabled and mention in changelog + ask on anyio/trio gitter and python discord (started hanging out a bit in the async channel) for people to try it out and give feedback if it should be default-enabled or not.
I should also add the linter to the trio/ repo

jakkdl:
after messing around in trio:

1. this is implemented in ruff preview as https://docs.astral.sh/ruff/rules/unused-async/
   * their tests are in https://github.com/astral-sh/ruff/blob/f4263490513ff1f5283b6801de4744a28f36df98/crates/ruff_linter/resources/test/fixtures/ruff/RUF029.py and include a couple cases I missed, e.g. `(x async for y in z)` 
2. We definitely need to suppress it for methods, cray false alarm rate. If one wanted to be very specific you could specifically look for `__aiter__` and stuff but it's probably not worth it. RUF029 is disabled for class methods
3. RUF029 further has lots of edge cases brought up in https://github.com/astral-sh/ruff/issues/10980, but we probably don't need to be as thorough.

tests/eval_files/async124.py
```diff
@@ -0,0 +1,143 @@
...
+async def default_value():
+    def foo(arg=await foo()): ...
+
+
+# 124 doesn't care if you evaluate the comprehension or not
+# 910 is stingy
+async def foo_async_gen():
+    return (  # ASYNC910: 4, "return", Statement("function definition", lineno-1)
+        a async for a in foo_gen()
+    )
```
Zac-HD, line 116: I think the behavior of 910 is clearly correct here; there are no checkpoints until after `foo_async_gen()` has returned, so that function could be sync.

Zac-HD (approved):
Looks good to me; optional nitpick below.

Zac-HD:
I'm also inclined to include FastAPI support in the default no-checkpoint-decorators; reducing the low-value-alarm rate for users is more important than my personal distaste for missing checkpoints 😅.  That can easily go in a future PR though.

tests/eval_files/async124.py
jakkdl: yeah, but the problem is differentiating that case from 
```python
k = (a async for a in foo_gen())
return list(k)
```
and similar. For 910 we can simply ignore the generator and tell the user to insert explicit checkpoints, but for a default-enabled check I think ASYNC124 should err on the side of expecting the generator to be consumed in the same function.

jakkdl:
> I'm also inclined to include FastAPI support in the default no-checkpoint-decorators; reducing the low-value-alarm rate for users is more important than my personal distaste for missing checkpoints 😅. That can easily go in a future PR though.

hrm. suppressing it for async124 makes lots of sense, but less so for async9xx... if they got suppressed for cases where async124 would get suppressed >_<
I guess that's a point in favor of default-disabling it for async124 in particular regardless of no-checkpoint-decorator

tests/eval_files/async124.py
Zac-HD: Snippet is invalid; list() doesn't do async iteration.  Which is a nitpick but relevant in that it means we probably don't need to distinguish such cases; consuming an async generator requires a syntactic checkpoint.

jakkdl: oooh, thank you!

jakkdl: Opened https://github.com/astral-sh/ruff/issues/14167 and https://github.com/python/mypy/issues/18124

I've now opened 7 issues today in mypy/ruff/pyright/typeshed :innocent:

jakkdl:
> > I'm also inclined to include FastAPI support in the default no-checkpoint-decorators; reducing the low-value-alarm rate for users is more important than my personal distaste for missing checkpoints 😅. That can easily go in a future PR though.
> 
> hrm. suppressing it for async124 makes lots of sense, but less so for async9xx... if they got suppressed for cases where async124 would get suppressed >_< I guess that's a point in favor of default-disabling it for async124 in particular regardless of no-checkpoint-decorator

opened #313 to procrastinate on this while I have way-too-many-balls in the air simultaneously :innocent:

jakkdl:
So this one fell by the wayside for.. no real reason. But I've merged main and done some minor edits, I'll merge it in a couple days unless somebody has any comments

PR #316, opened 2024-11-15
async101 & async119 now respects transform-async-generator-decorators

### LINKED ISSUE #315: `transform-async-generator-decorators` should affect `ASYNC101` as well as `ASYNC900`
since listing a decorator claims that it's safe, we should believe the user.

As it happens I just found a case where the transform decorator in https://github.com/python-trio/trio/issues/638 isn't fully safe (it can hang indefinitely on early exit)... I think the solution to that is to upstream a working version into Trio rather than continuing to have them proliferate with bugs, and so we'll presumably expand the default allowlist with that once (if?) my upcoming PR is merged.

jakkdl:
fixes #315 

I noticed the flag wasn't documented at all in `usage.rst`, so added it there.
I also think it should be enabled for ASYNC119?

The flag becomes somewhat weirdly named for ASYNC101 since it will also suppress errors in sync functions. Idk if it warrants a rename, or I should just mention it in the docs, or explicitly only make it have an effect on async funcs, or something else.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,12 +4,9 @@ Changelog
 
 `CalVer, YY.month.patch <https://calver.org/>`_
 
-FUTURE
-======
-- :ref:`ASYNC101 <async101>` and :ref:`ASYNC119 <async119>` are now silenced for decorators in :ref:`transform-async-generator-decorators`
-
 24.10.2
 =======
+- :ref:`ASYNC101 <async101>` and :ref:`ASYNC119 <async119>` are now silenced for decorators in :ref:`transform-async-generator-decorators`
 - :ref:`ASYNC102 <async102>` now also warns about ``await()`` inside ``__aexit__``.
 
 24.10.1
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,9 +4,9 @@ Changelog
 
 `CalVer, YY.month.patch <https://calver.org/>`_
 
-24.11.1
-=======
-- :ref:`ASYNC101` and :ref:`ASYNC119` are now silenced for decorators in :ref:`transform-async-generator-decorators`
+FUTURE
+======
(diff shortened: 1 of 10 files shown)


jakkdl:
hrm, I kinda wanna make the `test_changelog_and_version` tests be part of pre-commit. I keep forgetting to check those before pushing

docs/changelog.rst
```diff
@@ -4,6 +4,10 @@ Changelog
 
 `CalVer, YY.month.patch <https://calver.org/>`_
 
+24.11.1
+=======
+- :ref:`ASYNC101` and :ref:`ASYNC119` are now silenced for decorators in :ref:`transform-async-generator-decorators`
```
Zac-HD, line 9: probably remove the changelog from the first of this and #317 to merge, so we do a single release?

Zac-HD (approved):

jakkdl:
> hrm, I kinda wanna make the `test_changelog_and_version` tests be part of pre-commit. I keep forgetting to check those before pushing

wait, I just noticed we have
```python
def update_version() -> None:
    # If we've added a new version to the changelog, update __version__ to match
    last_version = next(iter(get_releases()))
    if last_version != VERSION:
        INIT_FILE = ROOT_PATH / "flake8_async" / "__init__.py"
        subs = (f'__version__ = "{VERSION}"', f'__version__ = "{last_version}"')
        INIT_FILE.write_text(INIT_FILE.read_text().replace(*subs))
```
which means the test for checking changelog matching `__version__` is kinda redundant (contributor docs also say you should increment `__version__`). It'd perhaps be safer to do `assert last_version == VERSION`, and if we want the convenience of replacing `__version__` that should perhaps be done in the pre-commit hook.

jakkdl:
Oh I can't even have `FUTURE` stuff in the changelog because tests must pass to merge. I'll def do some retooling of this at some point

PR #317, opened 2024-11-15
async100 now ignores trio.open_nursery and anyio.create_task_group

jakkdl:
I was cleaning up old branches in my fork, and it... seems like I never opened a PR for this? Wrote it several months ago, so decided to just rebase and clean it up

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,7 +5,7 @@ Changelog
 `CalVer, YY.month.patch <https://calver.org/>`_
 
 24.11.1
-======
+=======
 - :ref:`ASYNC100 <async100>` now ignores :func:`trio.open_nursery` and :func:`anyio.create_task_group` as cancellation sources.
 
 24.10.2
@@ -27,6 +27,7 @@ Changelog
 24.9.3
 ======
 - :ref:`ASYNC102 <async102>` and :ref:`ASYNC120 <async120>`:
+
   - handles nested cancel scopes
   - detects internal cancel scopes of nurseries as a way to shield&deadline
   - no longer treats :func:`trio.open_nursery` or :func:`anyio.create_task_group` as cancellation sources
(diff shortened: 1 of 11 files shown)


docs/changelog.rst
```diff
@@ -4,6 +4,10 @@ Changelog
 
 `CalVer, YY.month.patch <https://calver.org/>`_
 
+24.11.1
+======
+- :ref:`ASYNC100 <async100>` now ignores :func:`trio.open_nursery` and :func:`anyio.create_task_group` as cancellation sources.
```
Zac-HD, line 9:
```suggestion
- :ref:`ASYNC100 <async100>` now ignores :func:`trio.open_nursery` and :func:`anyio.create_task_group` as cancellation sources,
  because they are schedule points but not cancellation points.
```

This is a subtle enough concept that we should not assume users are familiar with it - I'd also call out this distinction in the docs for `ASYNC100`, and create glossary entries for both which are linked from "checkpoint" and link in turn to Trio docs (eg for `cancel_shielded_checkpoint()` and `checkpoint_if_cancelled()`)

Zac-HD (approved):
LGTM with some more documentation as below - thanks!

docs/changelog.rst
jakkdl: good call, feel free to reword my attempt :)

docs/changelog.rst
```diff
@@ -23,6 +27,7 @@ Changelog
 24.9.3
 ======
 - :ref:`ASYNC102 <async102>` and :ref:`ASYNC120 <async120>`:
+
   - handles nested cancel scopes
```
jakkdl, line 31: unrelated fix: this was not getting parsed as a nested bullet list w/o an empty line

Zac-HD (approved):
Added some tiny tweaks to the docs for cross-references.  Thanks again @jakkdl!

PR #321, opened 2024-11-17
improve changelog/release handling

jakkdl:
move changelog&version check to pre-commit and to a separate CI job that skips on skip-release label

Remaining quirks:
1. The job won't re-run automatically on label assignment. Claude is telling me I can get there by splitting up the workflows across files, I'll give that a try later.
2. A CHANGELOG run will fail after getting merged to main, because it loses the label. Not sure how to resolve that neatly; maybe just always pass `--allow-future` if it triggers on `push`.

The test file is kind of ugly now in that it both looks like a pytest file and doesn't. I should probably add some comments at the very least.

EDIT: actually, I shouldn't skip the CI job with the skip-release label - I should just make that pass `--allow-future` as well.
(also the flag is somewhat opaquely named, probs time to do a proper argparse solution).

It's also not great that skip-release doesn't *actually* skip a release attempt, afair we just rely on pushing-the-same-version-again to twine to not do anything.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,10 +62,22 @@ jobs:
             - name: Run tests
               run: python -m tox -e flake8_7 -- --onlyfuzz --no-cov -n auto
 
+    check_release:
+        runs-on: ubuntu-latest
+        strategy:
+            fail-fast: false
+        if: ${{ ! contains(github.event.issue.labels.*.name, 'skip-release') }}
+        steps:
+            - uses: actions/checkout@v4
+            - name: Set up Python 3
+              uses: actions/setup-python@v5
+            - name: Test changelog & version python tests/test_changelog_and_version.py --ensure-tag
+              run: python3 tests/check_changelog_and_version.py
+
     release:
         runs-on: ubuntu-latest
-        needs: [pyright, test]
-        if: github.repository == 'python-trio/flake8-async' &&  github.ref == 'refs/heads/main'
+        needs: [pyright, test, check_release]
+        if: github.repository == 'python-trio/flake8-async' && github.ref == 'refs/heads/main'
         steps:
             - uses: actions/checkout@v4
             - name: Set up Python 3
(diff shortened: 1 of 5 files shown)


jakkdl:
also it doesn't seem to work, at least not when re-running it after adding the label.

Zac-HD:
I think we've now hit my "not actually worth the trouble" threshold, let's stick with the status quo?

jakkdl:
The status quo *definitely* needs changing. I just realized that I forgot to undo the temporary changelog edit in https://github.com/python-trio/flake8-async/pull/316 when releasing https://github.com/python-trio/flake8-async/pull/317 (and I realized that... as I forgot to update the changelog in #323 before pushing). But I could look at making a simpler solution

jakkdl:
okay yeah there's a pretty simple way forward. Let's skip the label, and either have CI always allow `FUTURE` (but with a check so we don't release); or have CI always error on `FUTURE` (but it's tested in a separate job that's not required to merge, and main/ will error until we release). Let me know if you have any preferences between those two.

Will keep pre-commit, and fix `update_version()` that's been silently trying to edit `README.md` despite the example pre-commit config no longer being there.

.github/workflows/ci.yml
```diff
@@ -62,10 +62,21 @@ jobs:
...
+    check_release:
+        runs-on: ubuntu-latest
+        strategy:
+            fail-fast: false
+        steps:
+            - uses: actions/checkout@v4
+            - name: Set up Python 3
+              uses: actions/setup-python@v5
+            - name: Test changelog & version python tests/test_changelog_and_version.py --ensure-tag
+              run: python3 tests/check_changelog_and_version.py
```
Zac-HD, line 74: Name is inconsistent with command now?

Zac-HD (approved):
lgtm

.github/workflows/ci.yml
jakkdl: oops, looks like a bad copy-paste. fixed

PR #323, opened 2024-11-18
fix crash. Make identifier_to_string less footgunny.

### LINKED ISSUE #322: New crash since 24.9.5
```python
from contextlib import asynccontextmanager

@asynccontextmanager
async def fn():
    async with a().b():
        yield
```
crashes with
```python-traceback
...
  File ".../python3.11/site-packages/flake8_async/visitors/visitor91x.py", line 522, in visit_With_body
    self._checkpoint_with(node)
  File ".../python3.11/site-packages/flake8_async/visitors/visitor91x.py", line 511, in _checkpoint_with
    assert func is not None
           ^^^^^^^^^^^^^^^^
AssertionError
```
in 24.11.1; fine in 24.9.5

jakkdl:
fixes #322

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,20 +4,24 @@ Changelog
 
 `CalVer, YY.month.patch <https://calver.org/>`_
 
+24.11.2
+=======
+- Fix crash in ``Visitor91x`` on ``async with a().b():``.
+
 24.11.1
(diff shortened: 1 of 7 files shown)


Zac-HD (approved):
Thanks for the quick patch!

PR #344, opened 2025-01-31
make ASYNC912 and 913 care about cancel points, and ignore schedule points

### LINKED ISSUE #340: ASYNC913 should maybe catch `trio.open_nursery`
While not a deadlock issue (the loop will allow others to run), it's kind of silly to have an indefinite loop without a cancel point.

See https://github.com/python-trio/trio/issues/3192

jakkdl:
fixes #340 and extends it to ASYNC912 as well

I felt kinda silly realizing the `trio.open_nursery` <-> `anyio.create_task_group` replacement only now, that could clean up eval_files for... 100, 101, 105, 111, 112, 113 & 121 :sweat_smile:

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,10 @@ Changelog
 
 `CalVer, YY.month.patch <https://calver.org/>`_
 
+25.2.1
+=======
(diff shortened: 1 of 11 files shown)


jakkdl:
oh, derp, `trio.open_nursery` <-> `anyio.create_task_group` broke a bunch of eval files. I'll break that out into another PR

Zac-HD (approved):
Nice - this is a pretty subtle point, so catching the edge cases is particularly helpful!

PR #345, opened 2025-02-03
Improve replacement

jakkdl:
split out from #344 to clean up a lot of other eval_files to make use of the nursery/taskgroup replacing. Only recommended to review after merging #344

also fixes some async113 false alarms, so will need a version bump & changelog entry


Zac-HD (approved):
Nice!

PR #347, opened 2025-02-05
workaround bug generating invalid metadata

jakkdl:
https://github.com/python-trio/flake8-async/actions/runs/13154402924/job/36708351078#step:5:294 failed because of https://github.com/pypa/twine/issues/1216

Since the problematic actor apparently is setuptools I opted to fix it by setting `license_files`


jakkdl:
https://pypi.org/project/flake8-async/25.2.2/
:rocket: :tada:

PR #348, opened 2025-02-05
Make flake8 an optional dependency

jakkdl:
This has been lying around mostly-finished in a branch for a while, so I thought I'd clean it up and push it.

Writing our own parser of `noqa` was easy, but we still rely on `flake8` for parsing config files. This means a subset of users that:
a. use flake8 config
b. do not install any other flake8 plugins or flake8 itself in that venv, and
c. run the linter with `flake8-async`

will ~silently get a large change in behavior and they have to go look up the changelog to see they now need to install `flake8-async[flake8]`. If they don't do `c` they'll get a more obvious error that `flake8` is not available.
(only doing b and not c will also give a fairly obvious error)

I'm not sure how many people that will affect, but it might be worth quickly searching `pyproject.toml` or `.flake8` for the string `ASYNC` when `flake8-async` is invoked and give a big warning. Maybe even make it a two-step deprecation+release where we add the extra+warning now, and remove flake8 as a dependency in a couple weeks.


Zac-HD (approved):

jakkdl:
@Zac-HD no opinions on the ~breaking changes to running it?

Zac-HD:
I think few users are likely to be affected, but issuing the warning would be a nice upgrade.

jakkdl:
wait no I'm silly, if you're running `flake8-async` you've never had config parsing even if `flake8` is installed. So this will only impact people doing `pip install flake8-async && flake8 ...`
The warning could in theory still be helpful, but it'd raise false alarms for a bunch of legit usages, so I don't think it's a good idea.

Issue #173, opened 2023-03-28
Support match/case, exception groups


jakkdl:
I randomly realized TRIO91X (and maybe others) don't have any logic for match/case, and for any errors working with exceptions I should check that they also work with exception groups - I haven't checked how the AST looks for that.

jakkdl:
re exceptiongroups: Visitor123 defines visit_TryStar, but several others do not.

PR #358, opened 2025-03-05
add except* support to 102, 103, 104, 120, 910, 911 & 912

jakkdl:
fixes one half of #173 (the other being match/case)

First time trying out Claude Code, I more or less had to rewrite everything it generated in the end - but they gave it a good shot!

### DIFF
--- a/tests/eval_files/async91x_py311.py
+++ b/tests/eval_files/async91x_py311.py
@@ -1,0 +1,92 @@
+"""Test for ASYNC91x rules with except* blocks.
+
+ASYNC910: async-function-no-checkpoint
+ASYNC911: async-generator-no-checkpoint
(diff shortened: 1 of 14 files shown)


tests/eval_files/async91x_py311.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+async def foo_try_except_star_3():  # safe
+    try:
+        await foo()
+    except* ValueError:
+        raise
+
+
+# Multiple except* handlers - should all guarantee checkpoint
```
Zac-HD, line 46: checkpoint-or-raise?

Zac-HD (approved):
looking forward to this upgrade!

PR #363, opened 2025-03-31
Support structural pattern matching in ASYNC103,104 and 91X

### LINKED ISSUE #173: Support match/case, exception groups

jakkdl:
fixes #173

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,7 +42,7 @@ repos:
       hooks:
           - id: mypy
             # uses py311 syntax, mypy configured for py39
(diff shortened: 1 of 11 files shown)


Zac-HD (approved):
Looks great!  Let's merge tomorrow to match the release number to the month 😁

jakkdl:
haha, I didn't expect a same-day review :grin:

Zac-HD:
(definitely a good call this close to the line)

PR #366, opened 2025-04-02
add ASYNC125 constant-absolute-deadline

### LINKED ISSUE #353: New rule: do not pass `deadline=<constant>` to `CancelScope()`, `move_on_at()`, or `fail_at()`
These functions ([docs](https://trio.readthedocs.io/en/stable/reference-core.html#trio.CancelScope)) take a `deadline=` argument, which is defined relative to the current clock.  That in turn means that there's never a good reason to pass a constant, but it's pretty easy to write nonsense without realizing:

```python
with trio.fail_after(5): ...  # OK
with trio.fail_at(5): ...     # complete nonsense!
```

...but happily, it's also the kind of thing which is very easy to lint for.

jakkdl:
fixes #353 

anyio does not currently have `CancelScope(relative_deadline=...)`, but @agronholm seemed open to adding it so I'm not gonna bother messing around with the error message for now.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,10 @@ Changelog
 
 `CalVer, YY.month.patch <https://calver.org/>`_
 
+25.4.2
+======
+- Add :ref:`ASYNC125 <async125>` constant-absolute-deadline
+
 25.4.1
 ======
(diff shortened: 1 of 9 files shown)


Zac-HD (approved):
10/10, no notes.  Thanks @jakkdl, this will definitely catch some bugs!

PR #371, opened 2025-04-19
handle multiple withitems checkpointing each other

### LINKED ISSUE #367: False positives due to incorrect `async with` semantics
flake8-async seems to generally handle `async with` statements with multiple items incorrectly.

Below is an example demonstrating a false positive detection of `ASYNC100`, but this surely affects many other rules as well (e.g. `ASYNC912`)
```python
import asyncio

async def main():
    async with asyncio.timeout(1), asyncio.TaskGroup() as tg: # noqa: ASYNC100 ; false positive because tg.__aexit__ occurs within timeout scope
        _ = tg.create_task(asyncio.sleep(2))

asyncio.run(main())
```

jakkdl:
Also now handles autofixing async100 when there's multiple withitems, though it doesn't re-evaluate the With after autofixing so if a removed withitem was the only thing checkpointing another withitem, it doesn't clean up both.

fixes #367

also I have no clue how I totally hallucinated that `asyncio.timeouts.timeout` has ever been a thing??


tests/autofix_files/async100.py.diff
```diff
@@ -189,3 +189,41 @@
...
+-        trio.fail_after(5),  # error: 8, "trio", "fail_after"
+-        open("") as _,
+-        trio.move_on_after(5),  # error: 8, "trio", "move_on_after"
+-    ):
+-        ...
+-
+-    with (
+-        trio.move_on_after(10),  # error: 8, "trio", "move_on_after"
+         open("") as f,
+     ):
```
Zac-HD, line 228: Seems like a bug here, that it removes the non-cancel-scope context manager too in this case?

Zac-HD (approved):
one concern, otherwise looks good!

tests/autofix_files/async100.py.diff
jakkdl: ah no, that's the diff being dumb - an open is "added" on line 207. I'll add some comments to help the diff distinguish the `with`s

PR #374, opened 2025-04-29
add trio.as_safe_channel to async900 safe decorators

### LINKED ISSUE #370: suggest `@as_safe_channel` in ASYNC900
https://github.com/python-trio/trio/pull/3197
and async900 should also consider generators marked with the decorator as safe

jakkdl:
fixes #370 
I thought this was gonna be a one-line fix, but ended up getting really messy. There's definitely a cleaner way of implementing this, but it should work... :upside_down_face: 

The reason it gets messy is because `has_decorator`->`_get_identifier` only matches against the last part of an attribute, but I want `trio.as_safe_decorator` in the error message, so the message and the implementation gets a mismatch. The main alternative approach would be to have separate lists of decorators for the error message and the logic.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,10 @@ Changelog
 
 `CalVer, YY.month.patch <https://calver.org/>`_
 
+25.4.4
+======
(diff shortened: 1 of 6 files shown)


Zac-HD (approved):
Looks great!

PR #377, opened 2025-05-01
fix async113 false alarms

### LINKED ISSUE #263: reduce false alarms for ASYNC113
Currently this will raise ASYNC113, start-soon-in-aenter:
```python
@asynccontextmanager
async def foo():
        nursery.start_soon(bar)
    yield
```
This clearly is not susceptible to the problem of "task might not start before yielding" though, `bar` is not just guaranteed to start before the yield, it's even guaranteed to finish.

It should be fairly straightforward to hold potential async113 errors in `@asynccontextmanager`s, and only raise them if a yield is encountered later in the same nursery.

Another example:
```python
@asynccontextmanager
async def foo():
    # do stuff
    yield
        nursery.start_soon(bar)
```
Here we can entirely ignore any `start_soon`s after the `yield`, although that might get tripped up up by weird complicated logic with conditional yields. But it would also get handled by the logic of "only error if followed by yield in the same nursery" anyway, so should just go with that.

jakkdl:
fixes #263 

I came up with another false alarm that requires a bigger fix, so leaving that one as a tomorrow problem. It appears that 2.5 years is past my limit of fully remembering complicated details of the innards of this code base :sweat_smile:

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,10 @@ Changelog
 
 `CalVer, YY.month.patch <https://calver.org/>`_
 
+25.5.1
+======
+- Fixed :ref:`ASYNC113 <async113>` false alarms if the ``start_soon`` calls are in a nursery cm that was closed before the yield point.
+
 25.4.4
 ======
 - :ref:`ASYNC900 <async900>` now accepts and recommends :func:`trio.as_safe_channel`.
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -33,7 +33,7 @@ adding the following to your ``.pre-commit-config.yaml``:
    minimum_pre_commit_version: '2.9.0'
    repos:
    - repo: https://github.com/python-trio/flake8-async
-     rev: 25.4.4
+     rev: 25.5.1
      hooks:
        - id: flake8-async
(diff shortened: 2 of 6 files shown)


Zac-HD:
I wonder if it'd be easier to scan the body of the nursery block searching for `yield`s, rather than saving potential errors to emit later?  Small performance hit, but much easier to implement and I'd be less worried about subtle bugs.

Separately, I'd love to test on a nursery where we yield and _then_ start_soon a task.

jakkdl:
> I wonder if it'd be easier to scan the body of the nursery block searching for `yield`s, rather than saving potential errors to emit later? Small performance hit, but much easier to implement and I'd be less worried about subtle bugs.

eyy, this cracked it. And with setting `no_visit` if there's no `yield` I think this is a performance gain. *and* it fixed another false alarm, though one that relies on pretty egregious code.
...
and it let me remove the other code that led to other missed alarms since manually visiting doesn't let the `TypeTrackerVisitor` interject. bishbashbosh perfect fix

> Separately, I'd love to test on a nursery where we yield and _then_ start_soon a task.

:+1:

Zac-HD (approved):
needs changelog etc, then merge 🚀

PR #379, opened 2025-05-06
add async400 exceptiongroup-invalid-access

jakkdl:
pretty straightforward, bulk of it is in avoiding false alarms

### DIFF
--- a/flake8_async/visitors/visitor4xx.py
+++ b/flake8_async/visitors/visitor4xx.py
@@ -24,7 +24,9 @@ EXCGROUP_ATTRS = (
     # from BaseException
     "args",
     "with_traceback",
-    "add_notes",
+    "add_note",
+    # in the backport
+    "_is_protocol",
 )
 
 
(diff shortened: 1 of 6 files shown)


jakkdl:
hm, in hindsight I maybe should've used `TypeTrackerVisitor` - but it's a non-trivial rewrite so /shrug

flake8_async/visitors/visitor4xx.py
```diff
@@ -0,0 +1,96 @@
...
+    "message",
+    "exceptions",
+    "subgroup",
+    "split",
+    "derive",
+    # from BaseException
+    "args",
+    "with_traceback",
+    "add_notes",
+)
```
Zac-HD, line 28: Let's add a test that `set(dir(ExceptionGroup)).issubset(EXCGROUP_ATTRS)`, so that if a new method or attribute is added (e.g. by [PEP-785](https://peps.python.org/pep-0785/) 😅) we'll notice as soon as we test on that version of Python, rather than when users complain.

Zac-HD (approved):
one more test as below, then let's go!

flake8_async/visitors/visitor4xx.py
jakkdl: oopie, that caught a typo. Turns out it's `add_note` and not `add_notes` ^^;

PR #380, opened 2025-05-06
async102 & async120 no longer requires cancelscope to have a timeout

### LINKED ISSUE #272: ASYNC120: unintuitive error message
I recently bumped deps for trio-parallel and I got an ASYNC120:

https://github.com/richardsheridan/trio-parallel/actions/runs/9820324879/job/27114776180?pr=422#step:6:13

Which is here:

https://github.com/richardsheridan/trio-parallel/blob/7ccb4b75f36148952a6fec90bb94f3335dd8db2d/trio_parallel/_proc.py#L109-L117

Because this checkpoint is shielded, a `Cancelled` cannot be raised out of it to overwrite the exception.  I feel this is a bug of some kind because there are nearby shielded checkpoints in exceptions that don't raise this warning.

jakkdl:
fixes #272

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -6,7 +6,7 @@ Changelog
 
 25.5.2
 ======
-- :ref:`ASYNC120 <async120>` no longer requires cancel scopes to have a timeout. `(issue #272) <https://github.com/python-trio/flake8-async/issues/272>`_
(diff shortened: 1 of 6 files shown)


jakkdl:
2025-05-07T12:45:46,087293502+02:00
the test failure was some weird backend issue

jakkdl:
prettified all urls in the changelog, and also noticed that one of them was pointing at the entirely wrong issue (may have been copypasted from previous entry)

PR #385, opened 2025-05-23
start tagging releases again

jakkdl:
#321 intended to move changelog/version checks to pre-commit, but it *also* removed tagging releases.. without reintroducing that anywhere else.

shoutout to @A5rocks for noticing

We could perhaps do the tagging in pre-commit as well, but that would require contributors to remember to push with `--tag`.

We could go back and "manually" create tags for all past releases since #321, but I'm not sure there's much reason to? But if it's to be done it should be done now before we start to create new tags

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,4 +88,5 @@ jobs:
                   TWINE_USERNAME: __token__
                   TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
               run: |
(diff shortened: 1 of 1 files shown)


A5rocks:
> But if it's to be done it should be done now before we start to create new tags

Not necessarily, lightweight tags get marked as created as the same time as the commit they tag.

Zac-HD (approved):
I think it's worth keeping the `check_release` job, so that we get feedback on this before merging?  Otherwise LGTM

PR #408, opened 2025-09-26
GitHub Actions: Test on Python 3.14 release candidate 3

cclauss:
Python v3.14 -- October 7th
* https://www.python.org/download/pre-releases
* https://www.python.org/downloads/release/python-3140rc3
* https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#allow-pre-releases


cclauss:
Do we want to modify the test files to be compatible with Python 3.14, or do we want to update the linter to complain about test files that are incompatible with Python 3.14?

PR #445, opened 2026-04-21
Various small fixes

### LINKED ISSUE #268: run compile() on eval_files
and/or make a feature request in ruff to run compile(), or implement "await in lambda" check in specific.

See https://github.com/python-trio/flake8-async/pull/265#discussion_r1635570344

### LINKED ISSUE #310: Various minor things
1. I thought it was somewhat doomed to use ruff+flake8-async, but after way too long stumbled upon https://docs.astral.sh/ruff/settings/#lint_external We should probably suggest using this in docs.

2. flake8 `--per-file-ignores` works with our 5-character error codes... for now. We should add a test to see when they get around to running validation on it and break compatibility in another way.
   * I also started looking at adding an `--per-file-disable` to get around it and for running as standalone, and it should be fairly doable

3. should add an example on how to run flake8-async through flake8 in pre-commit.
4. fix visitor102: https://github.com/python-trio/flake8-async/pull/307/files#diff-4852409e79e4eb0f3736b0d3c6b6bfd8c2a14411b92416bc203a387dd6986829

### LINKED ISSUE #398: `ASYNC300` should allow `return asyncio.create_task(...)`
```python3
import asyncio

async def a():
    pass

def f():
    return asyncio.create_task(a())
```
```shell
$ flake8-async t.py
t.py:7:12: ASYNC300 asyncio.create_task() called without saving the result
```

This should be allowed, IMO

### LINKED ISSUE #422: Extra (unnecessary) 3.12 test run in CI
An extra 3.12 Python run happens here:

https://github.com/python-trio/flake8-async/blob/61ee52be056a7743530311e485958171d44f6935/.github/workflows/ci.yml#L47

CI logs where this happens:

https://github.com/python-trio/flake8-async/actions/runs/20048717457/job/57499857248?pr=420#step:6:30

This seemingly adds 2 minutes to CI times...

Zac-HD:
fixes https://github.com/python-trio/flake8-async/issues/268, fixes https://github.com/python-trio/flake8-async/issues/310, fixes https://github.com/python-trio/flake8-async/issues/398, fixes fixes https://github.com/python-trio/flake8-async/issues/422.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
             - name: Run tests with flake8
               run: python -m tox -e flake8
             - name: Run tests without flake8
-              run: python -m tox -e "" -- --no-cov
+              run: python -m tox -e no_flake8 -- --no-cov
 
     slow_tests:
         runs-on: ubuntu-latest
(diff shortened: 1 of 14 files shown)


jakkdl (commented):
`# NOCOMPILE` should be documented in https://flake8-async.readthedocs.io/en/latest/contributing.html#magic-markers

PR #448, opened 2026-04-22
Warn on `ExceptionGroup` subclasses without a `.derive()` method

### LINKED ISSUE #334: Warn on ExceptionGroup subclass without `derive`
While there's probably some usecase which makes sense for this, at most someone could just implement like so:

```py
class CustomEG(ExceptionGroup):
  def derive(self, exs):
    return ExceptionGroup(self.message, exs)
```

to get the original behavior (I think). From https://github.com/python-trio/trio/issues/3175

### LINKED ISSUE #446: ASYNC102 should allow aclose_forcefully
```
async def _run_handler(stream: StreamT, handler: Handler[StreamT]) -> None:
    try:
        await handler(stream)
    finally:
        await trio.aclose_forcefully(stream)  # E
```
this should be allowed, I think. It currently gets this error message:
```
src/trio/_highlevel_serve_listeners.py:36:9: ASYNC102 await inside try/finally on line 33 must have shielded cancel scope with a timeout.
```

but this relies on the implementation of `aclose_forcefully`, which is documented as cancelling immediately.

Zac-HD:
Warn when a class inherits from `ExceptionGroup` / `BaseExceptionGroup` but doesn't override `derive`. Without that override, `split`/`subgroup` (as used by nursery and TaskGroup implementations) silently produce plain `ExceptionGroup` instances instead of the custom subclass.  Fixes #334.

Also fixes https://github.com/python-trio/flake8-async/issues/446.


jakkdl (commented):
This should be an ASYNC4xx rule, as those are for `ExceptionGroup`s. (This was also mentioned in my comment in #334)

PR #467, opened 2026-07-06
Add ASYNC233 pathlib blocking call rule

### LINKED ISSUE #396: New rule: `ASYNC233` for blocking `pathlib` methods
Reserving the code now, so that the `ruff` implementation can go ahead: https://github.com/astral-sh/ruff/pull/19619

paranoa233:
Closes #396.

## Summary
- add ASYNC233 for blocking pathlib.Path I/O methods in async functions
- track pathlib Path values from annotations and common constructors
- document the new rule and add changelog coverage

## Test Plan
- `python -m pytest tests/test_flake8_async.py -k "ASYNC233 or ASYNC212"`
- `python -m pytest tests/test_messages_documented.py tests/test_all_visitors_imported.py`
- `python -m ruff check flake8_async tests docs`

Full `python -m pytest` on Windows: 1180 passed, 170 skipped; 2 existing CLI-output assertions fail because local subprocess output uses CRLF/path separators.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -6,6 +6,7 @@ Changelog
 
 26.6.1
 ======
+- Add :ref:`ASYNC233 <async233>` blocking-pathlib-call to detect blocking ``pathlib.Path`` I/O methods in async functions. `(issue #396) <https://github.com/python-trio/flake8-async/issues/396>`_
 - Add :ref:`ASYNC127 <async127>` unmaintained-httpx: use ``httpx2`` instead of ``httpx``, which is no longer maintained, to get security updates. `(issue #460) <https://github.com/python-trio/flake8-async/issues/460>`_
 - :ref:`ASYNC210 <async210>`, :ref:`ASYNC211 <async211>` and :ref:`ASYNC212 <async212>` now also detect blocking calls made through ``httpx2``, and their messages recommend ``httpx2.AsyncClient`` instead of ``httpx.AsyncClient``.
 
--- a/docs/rules.rst
+++ b/docs/rules.rst
@@ -190,6 +190,9 @@ ASYNC231 : blocking-fdopen-call
 ASYNC232 : blocking-file-call
     Blocking sync call on file object, wrap the file object in :func:`trio.wrap_file`/:func:`anyio.wrap_file` to get an async file object.
 
+ASYNC233 : blocking-pathlib-call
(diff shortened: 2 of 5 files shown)


docs/changelog.rst
```diff
@@ -6,6 +6,7 @@ Changelog
 
 26.6.1
 ======
+- Add :ref:`ASYNC233 <async233>` blocking-pathlib-call to detect blocking ``pathlib.Path`` I/O methods in async functions. `(issue #396) <https://github.com/python-trio/flake8-async/issues/396>`_
```
Zac-HD, line 9: new heading for this (26.7.2 - you might want to rebase), since it'll be a new release.

docs/rules.rst
```diff
@@ -190,6 +190,9 @@ ASYNC231 : blocking-fdopen-call
 ASYNC232 : blocking-file-call
     Blocking sync call on file object, wrap the file object in :func:`trio.wrap_file`/:func:`anyio.wrap_file` to get an async file object.
 
+_`ASYNC233` : blocking-pathlib-call
+    Blocking sync call to ``pathlib.Path`` I/O methods in async function, use :class:`trio.Path`/:class:`anyio.Path`. ``asyncio`` users should consider `aiopath <https://pypi.org/project/aiopath>`__ or `anyio`_.
+
 ASYNC240 : blocking-path-usage
     Avoid using :mod:`os.path` in async functions, prefer using :class:`trio.Path`/:class:`anyio.Path` objects. ``asyncio`` users should consider `aiopath <https://pypi.org/project/aiopath>`__ or `anyio`_.
```
Zac-HD, line 197: I think in each case I'd recommend that `asyncio` users use `anyio.Path` objects on asyncio backend (it's automatic and therefore easy), rather than aio- libs; I just can't trust the latter as much.

It's fine if you'd rather leave that for someone else's later PR though.

Zac-HD (commented):
Small docs and changelog note; otherwise looks great - thanks @paranoa233!

jakkdl (changes requested):
the implementation looks suspiciously big to me, when all the other `ASYNC23X` rules are very compact. It looks a bit like my early attempts at getting LLM's to help when models were less capable and couldn't properly juggle an AST and make use of the helpers properly.

It looks like it's parsing annotations, but there's a utility visitor for doing that.

So, I would like to hear the motivation for why all this plumbing is required - and if it even makes sense to mash it into `Visitor23X`.

